PR #4, opened 2024-05-24
feat: prompt for start block on rpc url networks in contract import

JasoonS:
Migrated PR from Float-Capital/indexer#1511

### DIFF
--- a/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
+++ b/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
@@ -184,6 +184,7 @@ impl ContractImportArgs {
 ///To validate that a user is not double selecting a network id
 fn prompt_for_network_id(
     opt_rpc_url: &Option<String>,
+    opt_start_block: &Option<u64>,
     already_selected_ids: Vec<u64>,
 ) -> Result<converters::NetworkKind> {
     //The first option of the list, funnels the user to enter a u64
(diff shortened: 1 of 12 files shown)


DenhamPreen:
https://github.com/Float-Capital/indexer/pull/1511/files @JonoPrest this addresses all the feedback from your review here. I am enabling automerge 🫡

codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
```diff
@@ -252,13 +254,41 @@ fn get_converter_network_u64(
...
+    .prompt();
+
+    let start_block: String = prompt.context("Failed during start block prompt")?;
+    let start_block: u64 = start_block
+        .trim()
+        .parse()
+        .context("Failed to parse start block, start block must be a number")?;
+
+    Ok(start_block)
+}
```
JonoPrest, line 290: @DenhamPreen have a look at inquires docs on "CustomType" https://github.com/mikaelmello/inquire?tab=readme-ov-file#customtype

It's not a big one but the parsing then becomes built in and on top of that you get an error message where you can retry rather than a failure during the prompting process.

JonoPrest (approved):
Thanks Den!

I've disabled auto-merge because there's one more comment I made on the actual prompt code. It's non blocking but can be an improvement to stop annoying exits half way through the import process.

PR #8, opened 2024-05-24
fix: use esbuild to reliably identify changes to the full/relevant import tree in eventhandler files

JasoonS:
Migrated PR from Float-Capital/indexer#1396


DenhamPreen:
I'm pretty disappointed that this PR was lost in the repo migration. I think it will be more work to try update this pr than to start this task a fresh. I don't have the capacity to do this right now, I will open a bug issue here and link to this.

DenhamPreen:
Closing due to staleness, new issue has been made https://github.com/enviodev/hyperindex/issues/277

JonoPrest:
Hey @DenhamPreen, sorry this never made it in. Looking at the changes it doesn't actually look too stale, shouldn't take much to back merge main in. I see this PR is tagged in the issue so we can relook at when we deal with it 👍🏼

PR #98, opened 2024-07-31
Perf optimisations

DZakh:


JonoPrest:
Hey @DZakh, I like the improvements and moving to static but looks like we still need to solve for the regression of adding round trips to the DB when they have already been requested in loaders.

Maybe as a suggestion instead of "idsNotInMemory" being an array. Make it a set, and on the response remove ids from the set as you initialize values. Then with the remaining set items initialize them as NotSet?

JonoPrest (approved):
Awesome stuff @DZakh looks great 👍🏼

PR #108, opened 2024-08-06
Add api token to template

JonoPrest:


codegenerator/integration_tests/src/main.rs
```diff
@@ -45,7 +45,7 @@ impl InitCombo {
             language: Some(l.clone()),
             init_commands: Some(init_flow),
             name: Some("test".to_string()),
-            api_token: None,
+            api_token: Some("4dc856dd-b0ea-494f-b27e-017b8b6b7e07".to_string()),
```
JonoPrest, line 48: Prompt should be skipped if it's added here in the arguments.

JonoPrest:
> Oh, is it what caused the out of memory error in integration tests?

No, Jason managed to clear the stale cache yesterday. It was a different failure where the interactive init was prompting for API key meaning it would halt the integration test.

PR #111, opened 2024-08-07
Preserve ordering of schema for docs

JonoPrest:


codegenerator/cli/npm/envio/evm.schema.json
```diff
@@ -11,15 +11,9 @@
...
-      "type": [
-        "array",
-        "null"
-      ],
-      "items": {
-        "$ref": "#/$defs/GlobalContract_for_ContractConfig"
-      }
+    "name": {
+      "description": "Name of the project",
+      "type": "string"
```
JonoPrest, line 16: Unfortunately I looked at the code for schema rs and it will always order the keyword "description" first for it's own objects so we basically just have a name collision there. I don't think it's very important to change right now. Not sure if we should even keep description in the schema.

DZakh: Hmmm, I think it's not used in any way.

DZakh (approved):

PR #115, opened 2024-08-12
Add transaction hash to config for contract import

JonoPrest:


moose-code (approved):
Amazing!!!

moose-code (commented):
The alternative is we don't include the transaction hash in the selection and we think of another way to name the the ID?

JonoPrest:
Closing in favour of #114

PR #119, opened 2024-08-13
Remove envio-win32-x64 as an optional dep

JonoPrest:
I discovered that trying to install envio with yarn, which will automatically install all peer deps and all optional deps. Yarn 🙄...

It breaks purely because the windows package doesn't exist. Lets remove it from the package.json until we are actually building for windows.

### DIFF
--- a/codegenerator/cli/npm/envio/package.json.tmpl
+++ b/codegenerator/cli/npm/envio/package.json.tmpl
@@ -39,8 +39,7 @@
     "envio-linux-x64": "${version}",
     "envio-linux-arm64": "${version}",
     "envio-darwin-x64": "${version}",
-    "envio-darwin-arm64": "${version}",
-    "envio-win32-x64": "${version}"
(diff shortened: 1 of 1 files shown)


DZakh (approved):

DZakh:
Might be worse updating `index.ts` as well, and handing the error for Windows users nicely

codegenerator/cli/npm/envio/package.json.tmpl
JasoonS, line 1: Thanks, good catch!

JasoonS (approved):

JonoPrest:
> Might be worse updating `index.ts` as well, and handing the error for Windows users nicely

I was thinking to do something but we already have:
```ts
throw new Error(`Couldn't find envio binary inside node_modules for ${os}-${arch}`)
```
Which I think is sufficient to cover any missing os/arch?

PR #128, opened 2024-08-19
Organize utils into namespaced modules

JonoPrest:
Utils felt like it needed some module namespacing...


DZakh (approved):
Nice, I also was thinking to do it

PR #129, opened 2024-08-20
fix: reorg default

DenhamPreen:


JonoPrest:
@DenhamPreen can you run `make update-schemas` from the cli folder and then tests should pass

DenhamPreen:
Thanks 👍

PR #132, opened 2024-08-20
Allow custom names for the events

DZakh:


codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -660,18 +655,30 @@ impl Event {
         human_cfg_event: EventConfig,
         opt_abi: &Option<EvmAbi>,
     ) -> Result<Self> {
-        let event = Event::get_abi_event(&human_cfg_event.event, opt_abi)?.into();
+        let event: NormalizedEthAbiEvent =
+            Event::get_abi_event(&human_cfg_event.event, opt_abi)?.into();
 
         Ok(Event {
+            name: human_cfg_event.name.unwrap_or(event.0.name.to_owned()),
```
DZakh, line 662: I'll add validation in another PR together with the fix for another reported bug where codegen fails with contract name containing dashes.

JonoPrest (commented):
I'm halfway through but will finish reviewing tomorrow morning 👍🏼

codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -358,14 +357,9 @@ pub struct EventTemplate {
 
 impl EventTemplate {
     pub fn from_config_event(config_event: &system_config::Event) -> Result<Self> {
-        let name = config_event
-            .get_event()
-            .name
-            .to_owned()
-            .to_capitalized_options();
+        let name = config_event.name.to_capitalized_options();
```
JonoPrest, line 360: I'm just trying to think if there is any place in the code where we need the orginal name of the event but I don't think so...

JonoPrest (approved):
Great implementation @DZakh, minimal changes, maximal effect 😄 🚀

PR #134, opened 2024-08-21
Fuel merge part 2 - include HyperFuelWorker into the hbs template

DZakh:
Also:
- Updated fuel greeter template to include `ecosystem` field, which will be mandatory after merge
- Added `fuel_greeter` test scenario - currently not working with local package version
- Removed the `event_decoder` field from system_config, so it's more generic and doesn't have ecosystem-specific dependencies.


scenarios/fuel_greeter/index.html
```diff
@@ -0,0 +1,70 @@
...
+    <title>GraphiQL</title>
+    <style>
+      body {
+        height: 100%;
+        margin: 0;
+        width: 100%;
+        overflow: hidden;
+      }
+
+      #graphiql {
```
JonoPrest, line 21: I think this is an artifact from the old fork. Maybe we remove this? It's just graphql playgroud alternative to hasura.

JonoPrest (approved):
Looks great Dmitry 👍🏼

PR #139, opened 2024-08-23
Fuel abi update to latest schema

JasoonS:
This seems to be mostly working - but is not well tested.

I was testing this with: `lenvio init fuel contract-import local -a ./exampleabi.json -n testing --contract-address 0x5b5e2102ab9b39e7f10928c5be1c31d2cfcd7555b0016ef4e2f41dca04323676 --contract-name "exampleTestContract" --api-token "123456" -l typescript -o ./test2 -d ./test2`

The abi file I was using to test is here: [Props721Edition-contract-abi.json](https://github.com/user-attachments/files/16726317/Props721Edition-contract-abi.json)

### DIFF
--- a/codegenerator/cli/src/fuel/abi.rs
+++ b/codegenerator/cli/src/fuel/abi.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Context, Result};
-use fuel_abi_types::abi::program::ProgramABI;
+use fuel_abi_types::abi::program::{ConcreteTypeId, MetadataTypeId, TypeId};
+use fuel_abi_types::abi::unified_program::{UnifiedProgramABI, UnifiedTypeApplication};
 use itertools::Itertools;
 use std::{collections::HashMap, fs, path::PathBuf};
 
@@ -10,7 +11,7 @@ use crate::rescript_types::{
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct FuelType {
-    pub id: usize,
+    pub id: TypeId,
     pub rescript_type_decl: RescriptTypeDecl,
     pub abi_type_field: String,
 }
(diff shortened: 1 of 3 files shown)


codegenerator/cli/src/fuel/abi.rs
```diff
@@ -59,48 +60,26 @@ pub struct Abi {
...
-        }
+    fn decode_types(program: &UnifiedProgramABI) -> Result<HashMap<TypeId, FuelType>> {
+        let mut types_map: HashMap<TypeId, FuelType> = HashMap::new();
 
-        fn mk_type_id_name(type_id: &usize) -> String {
-            format!("type_id_{}", type_id)
+        fn mk_type_id_name(type_id: &TypeId) -> String {
+            match type_id {
+                TypeId::Concrete(ConcreteTypeId(id)) => format!("concrete_type_id_{}", id),
+                TypeId::Metadata(MetadataTypeId(id)) => format!("metadata_type_id_{}", id),
```
JonoPrest, line 79: It would still be nice to make these meaningful names in future. I never got round to it on the first implementaion.

codegenerator/cli/src/fuel/abi.rs
```diff
@@ -112,214 +91,165 @@ impl Abi {
...
-                                        name: type_ident_name,
-                                        type_params,
-                                    }
-                                }
-                            };
-                            Ok((name, type_ident))
-                        })
-                        .collect::<Result<Vec<_>>>()
-                };
+            let type_id = TypeId::Metadata(MetadataTypeId(type_decl.type_id));
```
JonoPrest, line 99: Are all types in the program metadata? I seem to remember there were concrete types as well.

JonoPrest:
Hey @JasoonS, I'm not completely familiar with what's changed in fuel ABIs. 

There's a number of sections of code removed which I don't understand why or how its now handled. For eg. generics. Would you mind commenting your PR describing changes you made and why or hopping on a call with me and taking me through it?

JonoPrest:
Also, this code was pumped out pretty fast previously I think we should cover it in tests. It's hard for me to see what could break.

JasoonS:
Yes you're right it was also quite rushed, main success criteria was that it didn't crash on the example i gave it... - but we should definitely think how we can make this more robust.

JasoonS:
This can probably be simplified to how Dmitry did the indexer fuel codebase

DZakh:
Solved in the https://github.com/enviodev/hyperindex/pull/141

PR #140, opened 2024-08-23
Upgrade hs client to 0.6.2

JonoPrest:
Hs client < 0.6.2 took "Kind" as a field selection parameter for a transaction but would not serialize that into "type" for the actual query, so you would never get the value back. See this PR: https://github.com/enviodev/hypersync-client-node/pull/27


DZakh (approved):
Nothing to review here :) But it'd be nice to list how the upgrades affects the indexer.

JonoPrest:
> Nothing to review here :) But it'd be nice to list how the upgrades affects the indexer
True, I updated the description.

PR #142, opened 2024-08-23
Fix add proper handling for parent nodes in fetchstate

JonoPrest:
Unfortunately I made a mistake before with the TCO passing parent as args.

Previously, parent would be just the single node above the child. Now parent is essentially a reversed list that can accumulate and join with with a child register. Similar to how you can have a reverse accumulator for linked lists and then do a reverse concat.

I would like to add more tests regarding fetch state on Monday.


DZakh (approved):
I didn't process the logic of the PR. On the code level it looks good. Even if we missed something, I still think it's better than the current 2.1.1 😅

PR #148, opened 2024-08-27
Use envio@2.2.2-fuel which supports Address.toString in initial templates

DZakh:
It should fix integration tests in main


JonoPrest (approved):
👍🏼

PR #150, opened 2024-08-27
Fuel merge part 3 - wip

DZakh:


codegenerator/cli/templates/dynamic/codegen/src/Types.res.hbs
```diff
@@ -229,44 +229,11 @@ module {{contract.name.capitalized}} = {
...
-    }
-
-    let decodeHyperFuelData = (_) => Js.Exn.raiseError("HyperFuel decoder not implemented")
-    {{else}}
     @genType
-    type eventArgs = unit
-
-    let eventArgsSchema = S.literal(%raw(`null`))->S.variant((. _) => ())
-
-    external convertHyperSyncEventArgs: HyperSyncClient.Decoder.decodedEvent => eventArgs = "%identity"
```
DZakh, line 266: The only regression is that it's not runtime-free anymore, but everything else should work the same as before.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -791,10 +798,19 @@ impl Contract {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum EventPayload {
+    Params(Vec<EventParam>),
+    Data(RescriptTypeIdent),
```
JonoPrest, line 804: I'm pretty keen to use the same way of generating types for EVM in future. It's nice because it easily supports nested record types which we currently just inline as tuples.

DZakh: Sounds good, the only problem I see is building a hyperSyncConverter function

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -833,34 +849,56 @@ impl Event {
...
+            Some(log_id) => fuel_abi.get_log(&log_id)?,
+        };
 
-    pub fn get_event_inputs(&self) -> Vec<EventParam> {
-        self.get_event().inputs.clone()
+        Ok(Event {
+            name: event_config.name,
+            event,
+            payload: EventPayload::Data(log.data_type),
+            topic0: log.id,
```
JonoPrest, line 892: Is it appropriate to call fuel's log.id a sighash?

If so maybe we rename topic0 to sighash as a convention. Since topic0 is the sighash of an evm event.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -833,34 +849,56 @@ impl Event {
...
-        &self.event.0
-    }
+    // TODO: Validatate that all event names are unique
+    // TODO: Validatate that event names are not reserved
+    pub fn from_fuel_event_config(
+        event_config: FuelEventConfig,
+        fuel_abi: &FuelAbi,
+    ) -> Result<Self> {
+        let event: NormalizedEthAbiEvent =
+            Event::get_abi_event(&format!("{}()", event_config.name), &None)?.into();
```
JonoPrest, line 876: Cool so this is just a stub with an empty event for now? Need to relook at the data structure.

DZakh: Unfortunately it's still used in some places, so I couldn't get rid of it yet.

codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -349,55 +337,120 @@ impl EntityRecordTypeTemplate {
...
-                    is_eth_address: res_type == RescriptTypeIdent::Address,
-                    is_indexed: input.indexed,
-                    type_rescript_skar_decoded_param: res_type.to_string_decoded_skar(),
-                }
-            })
-            .collect::<Vec<_>>();
+    const DECODE_HYPER_FUEL_DATA_CODE: &'static str =
+        "(_) => Js.Exn.raiseError(\"HyperFuel decoder not implemented\")";
 
+    pub fn generate_convert_hyper_sync_event_args_code(params: &Vec<EventParam>) -> String {
```
JonoPrest, line 351: Can we handle the empty vec in this function so the context makes a bit more sense?

So check if it's empty and return the `(Utils.magic: HyperSyncClient.Decoder.decodedEvent => eventArgs)` here if so.

codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -349,55 +337,120 @@ impl EntityRecordTypeTemplate {
...
+                        params
+                            .iter()
+                            .map(|p| {
+                                RescriptRecordField::new(
+                                    p.name.to_string(),
+                                    abi_to_rescript_type(&p.into()),
+                                )
+                            })
+                            .collect(),
+                    );
```
JonoPrest, line 429: Lets remove the above if else check and then if params are empty just make unit type expression and the rest of the logic can remain the same.

codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -349,55 +337,120 @@ impl EntityRecordTypeTemplate {
...
+            EventPayload::Data(type_indent) => Ok(EventTemplate {
+                name,
+                params: vec![],
+                data_type: type_indent.to_string(),
+                data_schema_code: type_indent.to_rescript_schema(),
+                topic0: config_event.topic0.to_string(),
+                convert_hyper_sync_event_args_code:
+                    "(Utils.magic: HyperSyncClient.Decoder.decodedEvent => eventArgs)".to_string(),
+                decode_hyper_fuel_data_code: Self::DECODE_HYPER_FUEL_DATA_CODE.to_string(),
+            }),
```
JonoPrest, line 452: Actually this case can be covered as well. All we want to do on the match of these types is return a rescript type expression.

DZakh: It'll change in the future with the `decode_hyper_fuel_data_code` implementation.

codegenerator/cli/src/rescript_types.rs
```diff
@@ -576,6 +685,26 @@ mod tests {
...
+                RescriptRecordField::new("".to_string(), RescriptTypeIdent::Bool),
+                RescriptRecordField::new("1".to_string(), RescriptTypeIdent::Bool),
+                RescriptRecordField::new("Capitalized".to_string(), RescriptTypeIdent::Bool),
+                // Invalid characters in the middle are not handled correctly. Still keep the test to pin the behaviour.
+                RescriptRecordField::new("dashed-field".to_string(), RescriptTypeIdent::Bool),
+            ]),
+            vec![],
+        );
+
+        let expected = r#"type myRecord = {@as("module") module_: bool, @as("") _: bool, @as("1") _1: bool, @as("Capitalized") capitalized: bool, dashed-field: bool}"#.to_string();
```
JonoPrest, line 703: What's happening with dash-field? If you want, there's a lib I installed called convert_case which you could transform kebab to camel or snake case for eg.

codegenerator/cli/templates/dynamic/codegen/src/Types.res.hbs
JonoPrest: Doesn't the Utils.magic function become zero cost during optimization steps?

codegenerator/cli/src/rescript_types.rs
DZakh: It'll codegen like this `dashed-field: bool` and break during compilation. I think we don't have a case for this yet, so I just ignored it.

codegenerator/cli/src/hbs_templating/codegen_templates.rs
DZakh: Done

codegenerator/cli/src/config_parsing/system_config.rs
JonoPrest: Hmmm, yeah, but we should be able to lookup nested types and codegen a converter without too much trouble.

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: Done in the PR

codegenerator/cli/templates/dynamic/codegen/src/Types.res.hbs
DZakh: Only when the function is called from the same file

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: What about the `is_indexed`?

JonoPrest: > Should work, but what about the `is_indexed`?

Well, we should generate the converter from the EVM event, not the rescript type. If you add an index to a nested struct/tuple/array, the actual data doesn't get stored on chain. Just a keccak hash so we don't really have to handle params that are a nested index structure.

PR #153, opened 2024-08-28
Fix static address checksumming

DZakh:


codegenerator/cli/templates/dynamic/codegen/src/RegisterHandlers.res.hbs
```diff
@@ -34,7 +34,7 @@ let registerContractHandlers = (
           abi: Abis.{{contract.name.uncapitalized}}Abi->Ethers.makeAbi,
           addresses: [
             {{#each contract.addresses as | address |}}
-            {{#if is_evm_ecosystem}}
+            {{#if ../../../is_evm_ecosystem}}
```
DZakh, line 37: Shitty that it doesn't complain. Good that integration tests caught it.

JonoPrest: Yeah, I think once we have we have completely minimised the codegenerated files, we should remove the handlebars templates and just work with strings.

JonoPrest (approved):
Thanks Dmitry, good spot 👍🏼

PR #156, opened 2024-08-29
Added timestamp as a type for the indexer. One could save event.block…

moose-code:
….timestamp as a timestamp type as opposed to an interger for example


DZakh (approved):
Looks good

DZakh (changes requested):
But could you write some integration tests in the `test_codegen`

JonoPrest:
Hey @moose-code, I think the type already exists in Table.res so it probably would work out the box but I agree with Dmitry could we make sure there's a test for it?

JasoonS (commented):
Get some tests in ;)

I actually think we need a generic test that tests all types being read/saved to postgres - it will make it trivially easy to add a test like this if we ever add more types in the future.

codegenerator/cli/src/rescript_types.rs
```diff
@@ -335,6 +337,12 @@ impl RescriptTypeIdent {
             Self::String => "S.string".to_string(),
             Self::ID => "S.string".to_string(),
             Self::Bool => "S.bool".to_string(),
+            Self::Timestamp => {
+                // Don't use S.unknown, since it's not serializable to json
+                // In a nutshell, this is completely unsafe.
+                "S.json(~validate=false)->(Utils.magic: S.t<Js.Json.t> => S.t<Js.Date.t>)"
+                    .to_string()
+            }
```
JonoPrest, line 345: So we don't want to serialize because postgres client should manage date conversion?

DZakh: Yes, it does it for us

DZakh (approved):

PR #159, opened 2024-08-30
Use latest viem version

DZakh:
Using v1 on generated code, caused some of the users with viem v2 in the handlers get ts errors on the Hosted Service, since their version was overwritten. Unless we bundle the generated code, the issue might reappear in the future, but still I think it's a good option to keep our dependencies up to date. According to the changelog, there are no breaking changes affecting the generated code, but it might cause problems for users using viem v1 in the handlers.


JonoPrest (approved):
Cool sounds good 👌🏼

PR #162, opened 2024-08-30
Script to help update chains

JasoonS:
image

This makes it much easier to add/remove chains. This is an intermediate solution until we implement dotfiles or similar.


DZakh (approved):
Looks good

JonoPrest:
@DZakh I've moved it to the cli module. I think it will be helpful to add scripts/tools to the hidden "script" subcommand. Then it becomes easy to extend

codegenerator/cli/Makefile
```diff
@@ -1,14 +1,19 @@
...
-	cargo run -- print-config-json-schema fuel > ./npm/envio/fuel.schema.json
+	cargo run -- script print-config-json-schema fuel > ./npm/envio/fuel.schema.json
 
 # Target to update help
 update-help:
 	@echo "Updating help doc..."
-	cargo run -- print-cli-help-md > ./CommandLineHelp.md
+	cargo run -- script print-cli-help-md > ./CommandLineHelp.md
+
+# Target to update help
```
JonoPrest, line 13:
```suggestion
# Target to help find missing networks
```

DZakh (approved):
Good, I only don't like that it's a part of the published binary, but we can solve this later.

PR #164, opened 2024-08-30
Trim trailing end slashes (and fix formatting)

JasoonS:


codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -612,53 +612,56 @@ impl SyncSource {
...
-                  return Err(anyhow!("EE106: The HyperSync url \"{}\" is incorrect format. The HyperSync url needs to start with either http:// or https://", url));
-                }
-                Ok(Self::HypersyncConfig(HypersyncConfig { endpoint_url: url, is_client_decoder }))
+            } => {
+              if !validate_url(&url) {
+                return Err(anyhow!("EE106: The HyperSync url \"{}\" is incorrect format. The HyperSync url needs to start with either http:// or https://", url));
+              }
+              // Trim any trailing slashes from the URL
+              let trimmed_url = url.trim_end_matches('/').to_string();
+              Ok(Self::HypersyncConfig(HypersyncConfig { endpoint_url: trimmed_url, is_client_decoder }))
```
JasoonS, line 664: These 3 lines are the only changes, Everything else is just formatting (indentation was wrong)

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -1274,4 +1277,30 @@ mod test {
         assert_eq!(name_with_special_chars, expected_name_with_special_chars);
         assert_eq!(name_with_numbers, expected_name_with_numbers);
     }
+
+    #[test]
+    fn test_hypersync_url_trailing_slash_trimming() {
```
JasoonS, line 1282: Decided to add a small test for good measure

DZakh (changes requested):
Nitpick: In this case I'd replace the `validate_url` function with the `parse_url` one, which would either return None, or Some with trimmed URL. Also, looks like we'd like to have the same logic for rpc URLs, not only hypersync

DZakh (commented):
Since you're away, I've done the change myself. @JonoPrest Could you review, please

JonoPrest (approved):
LFG

DZakh (approved):

PR #172, opened 2024-09-04
Add validation for project name on input

JonoPrest:


codegenerator/cli/src/cli_args/interactive_init/validation.rs
```diff
@@ -77,12 +78,26 @@ impl<T> UniqueValueValidator<T> {
...
     } else {
         Ok(Validation::Valid)
     }
 }
 
+// must start with a letter
+// only contain lowercase letters, numbers or underscorse
+// must have a maximum length of 63 characters
+pub fn is_valid_project_name_validator(s: &str) -> Result<Validation, CustomUserError> {
+    if is_valid_postgres_db_name(s) && is_valid_npm_package_name(s) {
```
JonoPrest, line 91: Note we actually put the name into the generated package.json so it also should be a valid package name.

codegenerator/cli/src/cli_args/interactive_init/mod.rs
```diff
@@ -118,8 +118,9 @@ pub async fn prompt_missing_init_args(
         None => {
             // TODO: input validation for name
             Text::new("Name your indexer:")
-                .with_default("My Envio Indexer")
+                .with_default("my_envio_indexer")
```
JonoPrest, line 121: Changed the default since it was an invalid name

codegenerator/cli/src/cli_args/interactive_init/validation.rs
JonoPrest: @DZakh why does it need to be a valid postgres name? Is that for the hosted service?

DZakh: > Note we actually put the name into the generated package.json so it also should be a valid package name.

Not necessarily, it won't break anything if it's invalid there, unless it doesn't break JSON

DZakh: > why does it need to be a valid postgres name

I don't know. As I remember we use repo name for the projects

codegenerator/cli/src/cli_args/interactive_init/mod.rs
DZakh: I like dashes for package names :)

DZakh, line 121:
```suggestion
                .with_default("my-envio-indexer")
```

DZakh (approved):

DZakh:
I've done it a little bit differently in the PR https://github.com/enviodev/hyperindex/pull/184

codegenerator/cli/src/cli_args/interactive_init/mod.rs
JonoPrest: Unfortunately it fails postgres validation hence my question about that 👍🏼

DZakh: I removed the validation in https://github.com/enviodev/hyperindex/pull/184

PR #184, opened 2024-09-09
Remove validation for indexer name

DZakh:
We use it only for package.json which is not used for anything, so I just removed all validation and transformation logic for now.

### DIFF
--- a/codegenerator/cli/src/cli_args/interactive_init/mod.rs
+++ b/codegenerator/cli/src/cli_args/interactive_init/mod.rs
@@ -20,7 +20,7 @@ use std::str::FromStr;
 use strum;
 use strum::{Display, EnumIter, IntoEnumIterator};
 use validation::{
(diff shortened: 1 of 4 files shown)


codegenerator/cli/src/cli_args/interactive_init/mod.rs
```diff
@@ -118,7 +118,7 @@ pub async fn prompt_missing_init_args(
         None => {
             // TODO: input validation for name
             Text::new("Name your indexer:")
-                .with_default("My Envio Indexer")
+                .with_default("envio-indexer")
```
DZakh, line 121: Maybe even remove the step and use the directory name?

JonoPrest (approved):
Ok sweet, I'm happy for that 👍🏼

DZakh:
@JonoPrest I also removed the step from contract import 

image

JonoPrest (approved):
👌🏼

PR #186, opened 2024-09-09
Fuel merge part 8 - It runs 🎉

DZakh:
image


codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -501,7 +517,10 @@ let eventSignatures = [{}]"#,
                     ))?;
 
                 format!(
-                    "let abi = %raw(`null`)\n{}\n{}",
+                    "let abi = Fuel.transpileAbi(%raw(`require(\"../../{}\")`))\n{}\n{}",
```
JonoPrest, line 520: I'm always nervous to see hardcoded relative paths. Should this not take some sort of parameter that's been parsed. Like generated relative path?

codegenerator/cli/src/rescript_types.rs
```diff
@@ -138,17 +144,30 @@ impl RescriptTypeDecl {
...
                 .collect::<Vec<String>>()
                 .join(", ");
-            format!("({}) => {}", params, self.type_expr.to_rescript_schema())
+            let type_name = format!(
+                "{}<{}>",
+                type_name,
+                self.parameters
+                    .iter()
+                    .map(|param| format!("'{}", param))
+                    .collect::<Vec<String>>()
```
JonoPrest, line 163: Think this collect should be unnecessary
```suggestion
```

codegenerator/cli/src/rescript_types.rs
```diff
@@ -197,19 +216,20 @@ impl RescriptTypeExpr {
...
                     .iter()
                     .map(|item| {
                         format!(
-                            r#"S.object(s =>
+                            r#"S.object((s): {} =>
 {{
   s.tag("case", "{}")
   {}({{payload: s.field("payload", {})}})
 }})"#,
+                            type_name,
```
JonoPrest, line 232: I think named parameters might be a bit easier to read with these longer string interpolations:

https://doc.rust-lang.org/std/fmt/#named-parameters

Also you don't have to pass the same value in twice like item.name here

codegenerator/cli/src/rescript_types.rs
```diff
@@ -744,14 +754,14 @@ mod tests {
                 RescriptRecordField::new("fieldA".to_string(), RescriptTypeIdent::Int),
                 RescriptRecordField::new("fieldB".to_string(), RescriptTypeIdent::Bool),
             ])
-            .to_rescript_schema(),
-            "S.object(s => {fieldA: s.field(\"fieldA\", S.int), fieldB: s.field(\"fieldB\", \
+            .to_rescript_schema(&"eventArgs".to_string()),
+            "S.object((s): eventArgs => {fieldA: s.field(\"fieldA\", S.int), fieldB: s.field(\"fieldB\", \
```
JonoPrest, line 758: What is the error you get if you don't have the type annotation? Does rescript not infer the type if it's in the same scope?

codegenerator/cli/templates/dynamic/codegen/src/ContextEnv.res.hbs
```diff
@@ -59,10 +59,12 @@ let getAddedDynamicContractRegistrations = (contextEnv: t<'eventArgs>) =>
 let makeDynamicContractRegisterFn = (~contextEnv: t<'eventArgs>, ~contractName, ~inMemoryStore) => (
   contractAddress: Address.t,
 ) => {
-  // Even though it's the Address.t on ReScript side, for TS side it's a string.
+  {{#if is_evm_ecosystem}} {{!-- TODO: Add validation for Fuel --}}
+  // Even though it's the Address.t type on ReScript side, for TS side it's a string.
   // So we need to ensure that it's a valid checksummed address.
   let contractAddress = contractAddress->Address.Evm.fromAddressOrThrow
 
+  {{/if}}
```
JonoPrest, line 67: Will you add a similar sanitisation function for fuel

codegenerator/cli/src/rescript_types.rs
DZakh: There are cases when there are multiple types with the same shape, and ReScript doesn't know which one to use

codegenerator/cli/templates/dynamic/codegen/src/ContextEnv.res.hbs
DZakh: Yes, I'll add it

DZakh:
Thanks for the review. I've fixed the code comments, and I'll do other fixes in the next PR

codegenerator/cli/src/hbs_templating/codegen_templates.rs
JonoPrest: I'm pretty sure even our integration tests would have this case.

DZakh: This is how `indexer-fuel` works. I'll cut the corner here, to make it work and then fix some of the issues. I already have 4 similar things on my list

JonoPrest (approved):
Cool, looks good Dmitry although I don't think the paths are handled properly for referencing the abi file. I would prefer we handle it rather than wait for it to come up.

PR #196, opened 2024-09-13
Fix raw events serializing

DZakh:


codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -492,7 +492,8 @@ impl EventTemplate {
                     name,
                     params: template_params,
                     data_type: data_type_expr.to_string(),
-                    data_schema_code: data_type_expr.to_rescript_schema(&"eventArgs".to_string()),
+                    params_raw_event_schema: data_type_expr
+                        .to_rescript_schema(&"eventArgs".to_string()),
```
DZakh, line 496: I assume this will also fail for events with empty params

`null value in column \"params\" of relation \"raw_events\" violates not-null constraint`

JonoPrest (approved):
Ok awesome thanks for the fix Dmitry 🙏🏼

PR #208, opened 2024-09-17
Fix second part of Fuel id on contract import

DZakh:
I've decided to keep it as before, since I'm not sure how logIndex behaves for Fuel. Can change it later to block number , but it should be `block.height` instead anyways.


JonoPrest (approved):
Ok sure 👍🏼

PR #209, opened 2024-09-17
Use block height for id on contract import

DZakh:


codegenerator/cli/src/hbs_templating/contract_import_templates.rs
```diff
@@ -281,13 +281,13 @@ impl Event {
         }
         .to_string();
 
-        let second_part = match is_fuel {
-            true => format!("${{{event_var_name}.transaction.id}}"),
-            false => format!("${{{event_var_name}.block.number}}"),
+        let block_number_field = match is_fuel {
+            true => "heigth",
```
JonoPrest, line 285: I think height is spelled wrong here.

DZakh:
Fixed and tested 🙏

codegenerator/cli/src/hbs_templating/contract_import_templates.rs
```diff
@@ -281,13 +281,13 @@ impl Event {
         }
         .to_string();
 
-        let second_part = match is_fuel {
-            true => format!("${{{event_var_name}.transaction.id}}"),
-            false => format!("${{{event_var_name}.block.number}}"),
+        let block_number_field = match is_fuel {
+            true => "height",
+            false => "humber",
```
JonoPrest, line 286: I'm pretty sure this should be number?
```suggestion
            false => "number",
```

JonoPrest (changes requested):
Sorry looks like another one...

DZakh:
I'm surprised by my ability making silly mistakes when I'm lazy to test something because I'm 100% sure that it works

JonoPrest (approved):
Ok cool, looks great 👌🏼

PR #216, opened 2024-09-19
Add benchmarks for diagnostics

JonoPrest:
Adds a couple more data points for useful diagnostics

Screenshot 2024-09-19 at 11 31 09


DZakh (approved):
I like the rounding with precision 2

PR #217, opened 2024-09-19
Add helper for running and printing benchmarks

JonoPrest:
Now you can run: 

`envio start --bench`

to run the indexer with benchmarks

and 

`envio benchmark-summary`

to get he print out.


DZakh (approved):

DZakh:
Having it as `-b` would be convinient

PR #225, opened 2024-09-24
Partial support for Fuel Bytes type

DZakh:
Treat it as unknown for now

### DIFF
--- a/scenarios/fuel_test/src/AllEventsHandlers.ts
+++ b/scenarios/fuel_test/src/AllEventsHandlers.ts
@@ -200,7 +200,6 @@ AllEvents.Option2.handler(async ({ event }) => {
 const vecLogSchema = S.array(SExtra.bigint);
 AllEvents.VecLog.handler(async ({ event }) => {
   vecLogSchema.assert(event.params)!;
-  console.log(event);
   expectType<AssertSchemaType<typeof event.params, typeof vecLogSchema>>(true);
 });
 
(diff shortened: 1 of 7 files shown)


scenarios/fuel_test/src/AllEventsHandlers.ts
```diff
@@ -196,7 +198,15 @@ AllEvents.Option2.handler(async ({ event }) => {
...
-AllEvents.Vec.handler(async ({ event }) => {
+AllEvents.VecLog.handler(async ({ event }) => {
   vecLogSchema.assert(event.params)!;
   expectType<AssertSchemaType<typeof event.params, typeof vecLogSchema>>(true);
 });
+
+const bytesLogSchema = S.unknown;
+AllEvents.BytesLog.handler(async ({ event }) => {
+  bytesLogSchema.assert(event.params)!;
+  expectType<AssertSchemaType<typeof event.params, typeof bytesLogSchema>>(
```
JasoonS, line 209: I think I need a little bit more explanation on this. So it's seen as an unknown type.

Does that mean it's the data exactly raw as is from the chain and it's up to the user to parse this type into whatever they want?

JasoonS (approved):
If I understand this change correctly, it doesn't change any functionality. It just explicitly marks the type of the data to unknown. And then the user can do whatever they want with that data, but we don't pause it for them.

scenarios/fuel_test/src/AllEventsHandlers.ts
DZakh: > Does that mean it's the data exactly raw as is from the chain and it's up to the user to parse this type into whatever they want?

Yes

DZakh:
> If I understand this change correctly, it doesn't change any functionality. It just explicitly marks the type of the data to unknown. And then the user can do whatever they want with that data, but we don't pause it for them.

Yes, previously it was treated as an object with some weird fields, which wasn't correct and broke codegen for schemas. In this version, we still don't have an exact type for bytes, but at least it's not lying.

PR #231, opened 2024-09-29
 Update feature for precision to use a config directive so that it can be consistent even for other fields going into the future.

JasoonS:
https://github.com/enviodev/docs/pull/488 - this was also updated in the docs. Lets get this in a release ASAP?

### DIFF
--- a/codegenerator/cli/src/config_parsing/entity_parsing.rs
+++ b/codegenerator/cli/src/config_parsing/entity_parsing.rs
@@ -528,13 +528,15 @@ impl Entity {
 fn get_positive_integer(arg_value: &Value<String>) -> anyhow::Result<u32> {
     match arg_value {
         Value::Int(i) => {
-            let val = i.as_i64().context("Failed to convert value to i64")?;
+            let val = i
+                .as_i64()
+                .context("EE217: Failed to convert value to i64")?;
             if val < 0 {
-                return Err(anyhow!("Value must be a positive integer"));
+                return Err(anyhow!("EE217: Value must be a positive integer"));
             }
             Ok(val as u32)
         }
-        _ => Err(anyhow!("Value must be an integer")),
+        _ => Err(anyhow!("EE217: Value must be an integer")),
     }
 }
 
@@ -546,7 +548,7 @@ pub struct Field {
 
 impl Field {
     fn from_obj_field(field: &ObjField<String>) -> anyhow::Result<Self> {
-        // Get all gql directives labeled @derivedFrom and @index
+        // Collect directives
         let derived_from_directives = field
             .directives
             .iter()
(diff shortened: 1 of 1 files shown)


codegenerator/cli/src/config_parsing/entity_parsing.rs
```diff
@@ -2020,19 +2007,19 @@ type TestEntity {
...
+        exampleBigInt: BigInt @config(precision: 76)
+        exampleBigIntRequired: BigInt! @config(precision: 77)
+        exampleBigIntArray: [BigInt!] @config(precision: 78)
+        exampleBigIntArrayRequired: [BigInt!]! @config(precision: 79)
+        exampleBigDecimal: BigDecimal @config(precision: 80, scale: 5)
+        exampleBigDecimalRequired: BigDecimal! @config(precision: 81, scale: 5)
+        exampleBigDecimalArray: [BigDecimal!] @config(precision: 82, scale: 5)
+        exampleBigDecimalArrayRequired: [BigDecimal!]! @config(precision: 83, scale: 5)
+        exampleBigDecimalOtherOrder: BigDecimal! @config(scale: 6, precision: 84)
     }
```
JasoonS, line 2023: This is a tldr of this PR

codegenerator/cli/src/config_parsing/entity_parsing.rs
```diff
@@ -2263,17 +2250,15 @@ type TestEntity {
 
         assert!(result.is_err());
         let err_message = format!("{:?}", result.unwrap_err());
-        assert!(err_message.contains(
-            "The numeric directive on a field is only suitable for BigDecimal scalar type."
-        ));
+        assert!(err_message.contains("EE215"));
```
JasoonS, line 2253: I think it is best just to match on the error string

JasoonS: Rather than longer messages that could be more brittle

codegenerator/cli/src/config_parsing/entity_parsing.rs
```diff
@@ -636,95 +617,101 @@ impl Field {
...
-                    "precision" => {
-                        precision = Some(get_positive_integer(arg_value).context(format!(
-                            "Parsing numeric.precision directive on BigDecimal with field name {}",
-                            &field.name
-                        ))?);
+        let mut pg_type_modifications = PgTypeModifications::default();
+
+        // Process @config
+        if let Some(config_directive) = config_directives.first() {
+            match underlying_scalar {
```
JasoonS, line 636: Basically the same logic, just put inside this match

JonoPrest:
@JasoonS something looks a bit off I think this needs to be rebased onto your base branch?

codegenerator/cli/src/config_parsing/entity_parsing.rs
```diff
@@ -636,95 +617,101 @@ impl Field {
...
-                    field.name
-                ));
-            }
-            let precision = get_positive_integer(arg_value).context(format!(
-                "Parsing precision.digits directive on BigInt field with field name {}",
-                &field.name
-            ))?;
-            pg_type_modifications.big_int_precision = Some(precision);
-        }
+        let underlying_scalar = field_type.get_underlying_scalar();
```
JonoPrest, line 630: Yeah I think just combine these. It doesn't make sense to have an incomplete version of "field_type" in scope unless it's needed.

JonoPrest (approved):
🚀  power

PR #234, opened 2024-09-30
Move from generated to package

DZakh:
This is the first step to a more library-like design with minimum generated code.


DZakh:
> But if we want to move the rest of the code, we'll need to add dependencies to the `envio` package. There's nothing bad in it, but they'll be listed on the npm page.

And I definitely want to move the rest of the code. It should boost DX, by allowing to run compiler and tests directly on the code we are working on. Also, it'll enforce a more modular design, since it won't allow dependency on the generated code.

JonoPrest (approved):
🕺🏼

PR #236, opened 2024-09-30
Clean cache guide

JasoonS:
image

We could definitely turn this into a better script for now, but since it is such a rare occurrence I think this is fine.


JonoPrest (approved):
🚀

PR #241, opened 2024-10-01
Improve contract import

DZakh:
- Adds `Transfer`, `Mint`, `Burn`, `Call` events in the selection for the Fuel contract import
- Fix broken schema.graphql - wasn't released (this part is not tested, so we'd miss it if I didn't test it manually)
- Remove the second linebreak between handlers in contract import
- Remove unused code for contract import codegen

### DIFF
--- a/codegenerator/cli/src/cli_args/interactive_init/fuel_prompts.rs
+++ b/codegenerator/cli/src/cli_args/interactive_init/fuel_prompts.rs
@@ -5,7 +5,8 @@ use crate::{
         ContractImportArgs, InitFlow as ClapInitFlow, LocalImportArgs, LocalOrExplorerImport,
         TemplateArgs,
     },
-    fuel::abi::{FuelAbi, FuelLog},
+    config_parsing::human_config::fuel::EventConfig,
+    fuel::abi::{FuelAbi, BURN_EVENT_NAME, CALL_EVENT_NAME, MINT_EVENT_NAME, TRANSFER_EVENT_NAME},
     init_config::fuel::{ContractImportSelection, InitFlow, SelectedContract, Template},
 };
 use anyhow::{Context, Result};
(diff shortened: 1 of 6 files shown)


codegenerator/cli/src/hbs_templating/contract_import_templates.rs
```diff
@@ -212,21 +212,6 @@ pub struct AutoSchemaHandlerTemplate {
...
-            .imported_contracts
-            .into_iter()
-            .map(|c| c.try_into())
-            .collect::<anyhow::Result<Vec<Schema>>>()?
-            .into_iter()
-            .flat_map(|s| s.entities.clone().into_values())
-            .collect::<Vec<_>>();
-        Schema::new(entities, vec![])
-    }
-}
```
DZakh, line 228: This is unused

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -969,10 +969,10 @@ impl Event {
...
                     } else {
                         match event_config.name.as_str() {
-                            "Mint" => EventType::Mint,
-                            "Burn" => EventType::Burn,
-                            "Transfer" => EventType::Transfer,
-                            "Call" => EventType::Call,
+                            MINT_EVENT_NAME => EventType::Mint,
+                            BURN_EVENT_NAME => EventType::Burn,
+                            TRANSFER_EVENT_NAME => EventType::Transfer,
+                            CALL_EVENT_NAME => EventType::Call,
```
JonoPrest, line 975: Isn't it possible for these to be aliased?

codegenerator/cli/src/cli_args/interactive_init/fuel_prompts.rs
```diff
@@ -120,9 +122,37 @@ async fn get_contract_import_selection(args: ContractImportArgs) -> Result<Selec
...
+    selected_events.push(EventConfig {
+        name: BURN_EVENT_NAME.to_string(),
+        log_id: None,
+        type_: None,
+    });
+    selected_events.push(EventConfig {
+        name: CALL_EVENT_NAME.to_string(),
+        log_id: None,
+        type_: None,
+    });
```
JonoPrest, line 153:
```suggestion
  let mut selected_events: Vec<EventConfig> = abi
      .get_logs()
      .iter()
      .map(|log| EventConfig {
          name: log.event_name.clone(),
          log_id: Some(log.id.clone()),
          type_: None,
      })
      .collect();
  
  let event_names = [
      TRANSFER_EVENT_NAME,
      MINT_EVENT_NAME,
      BURN_EVENT_NAME,
      CALL_EVENT_NAME,
  ];
  
  selected_events.extend(event_names.iter().map(|&name| EventConfig {
      name: name.to_string(),
      log_id: None,
      type_: None,
  }));
```

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: Possible, this is for the case when `type` is not specified and we try to derive it from name

DZakh: If they are aliased, then the `type` field should be used and we won't enter the match branch

PR #242, opened 2024-10-01
Benchmark Improvements

JonoPrest:
- Fixes benchmarking fo ts projects (benchmark was importing config from register handlers which required calling with ts-node)
- Fixes schema for null fetch stats
- Changes benchmark data structure from records to dicts and makes it more easy to add arbitrary benchmarks


DZakh (approved):
Like it

PR #243, opened 2024-10-01
Fix require for the bin.js file

DZakh:
`require.resolve` didn't work correctly with esm.

I get:

image


JonoPrest (approved):
Will this work when using a ts-node indexer?

PR #244, opened 2024-10-01
Fixes contract import with `0.0.0-main-...` versions

DZakh:
The path to envio is broken when installing the `0.0.0-main-...` package from npm. It adds a lousy fix for it:

```
"envio": "/Users/dzakh/.npm/_npx/ee52c6fd71a620c9/node_modules/envio-darwin-arm64/bin/envio/../../../cli/npm/envio"
```


JonoPrest (approved):
Good catch

PR #256, opened 2024-10-08
Fix copy function call happening on non reorg indexer

JonoPrest:
Unfortunately this was being triggered regardless of whether or not it was a rollback on reorg indexer.


JasoonS (approved):
Makes sense :+1:

PR #258, opened 2024-10-10
Fix restart bug with dynamic contracts

JonoPrest:
Dynamic contract restart resistance has been broken since we removed `raw_events` being required.

- Fixes the query that gets dynamic contract registrations on restart
- Adds test to assert the behaviour of a restart

### DIFF
--- a/codegenerator/cli/templates/static/codegen/src/db/DbFunctionsImplementation.js
+++ b/codegenerator/cli/templates/static/codegen/src/db/DbFunctionsImplementation.js
@@ -417,16 +417,14 @@ module.exports.deleteStaleEndOfBlockRangeScannedDataForChain = (
     ;`;
 };
 
-module.exports.readDynamicContractsOnChainIdAtOrBeforeBlock = (
+module.exports.readDynamicContractsOnChainIdBeforeEventId = (
   sql,
   chainId,
-  block_number,
+  eventId,
 ) => sql`
-  SELECT c.contract_address, c.contract_type, c.event_id, c.chain_id, c.block_timestamp
-  FROM "public"."dynamic_contract_registry" as c
-  JOIN raw_events e ON c.chain_id = e.chain_id
-  AND c.event_id = e.event_id
-  WHERE e.block_number <= ${block_number} AND e.chain_id = ${chainId};`;
+  SELECT *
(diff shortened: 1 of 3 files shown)


codegenerator/cli/templates/static/codegen/src/db/DbFunctionsImplementation.js
```diff
@@ -417,16 +417,14 @@ module.exports.deleteStaleEndOfBlockRangeScannedDataForChain = (
...
   sql,
   chainId,
-  block_number,
+  eventId,
 ) => sql`
-  SELECT c.contract_address, c.contract_type, c.event_id, c.chain_id, c.block_timestamp
-  FROM "public"."dynamic_contract_registry" as c
-  JOIN raw_events e ON c.chain_id = e.chain_id
-  AND c.event_id = e.event_id
-  WHERE e.block_number <= ${block_number} AND e.chain_id = ${chainId};`;
```
JonoPrest, line 429: There was a dependency on raw_events table for this start up query

codegenerator/cli/templates/static/codegen/src/db/DbFunctionsImplementation.js
```diff
@@ -417,16 +417,14 @@ module.exports.deleteStaleEndOfBlockRangeScannedDataForChain = (
...
+  SELECT *
+  FROM "public"."dynamic_contract_registry"
+  WHERE event_id < ${eventId} AND chain_id = ${chainId};`;
```
JonoPrest, line 427: Now I construct the eventId of the next block and filter for anything below that ID so you end up with the desired effect

codegenerator/cli/templates/static/codegen/src/db/DbFunctionsImplementation.js
JasoonS: I got to you so it would lose those dynamic contracts on restart.

JasoonS (approved):
Cool, it makes sense I haven't run it but the SQL query definitely looks much better being simplified like that :+1:

JonoPrest:
> Cool, it makes sense I haven't run it but the SQL query definitely looks much better being simplified like that 👍

The SQL gets run and validated in the test when it calls `makeFromDbState`

PR #273, opened 2024-10-18
Move ChainMap to npm package

DZakh:


codegenerator/cli/npm/envio/package.json.tmpl
```diff
@@ -28,6 +28,9 @@
     "envio-darwin-x64": "${version}",
     "envio-darwin-arm64": "${version}"
   },
+  "dependencies": {
+    "rescript": "11.1.3"
+  },
```
DZakh, line 33: Had to add the dependency because of the error https://github.com/enviodev/hyperindex/actions/runs/11438088469/job/31820281579

DZakh:
Need to change the integration tests to use the packed package instead of referencing the `../../codegenerator/cli/npm/envio`

codegenerator/cli/src/commands.rs
```diff
@@ -133,21 +133,6 @@ pub mod codegen {
         Ok(exit3)
     }
 
-    pub async fn npx_codegen(
-        envio_version: String,
-        project_paths: &ParsedProjectPaths,
-    ) -> anyhow::Result<()> {
-        let package = format!("envio@{}", envio_version);
-        let args = vec!["--yes", package.as_str(), "codegen"];
-        execute_command("npx", args, &project_paths.project_root)
```
DZakh, line 142: Thanks to the new local-bin.mjs setup managed to find this old code, where we ran fuel codegen via npx 😅

codegenerator/cli/src/executor/init.rs
```diff
@@ -236,11 +236,10 @@ pub async fn run_init_args(init_args: InitArgs, project_paths: &ProjectPaths) ->
...
             // Else install the local version for development and testing
             match env::current_exe() {
-                // This should be something like "~/envio/hyperindex/codegenerator/target/debug/envio" or ".../target/debug/integration_tests"
-                Ok(exe_path) => exe_path
-                    .join("../../../cli/npm/envio")
-                    .to_string_lossy()
-                    .to_string(),
+                // This should be something like "file:~/envio/hyperindex/codegenerator/target/debug/envio" or "file:.../target/debug/integration_tests"
+                Ok(exe_path) => {
+                    format!("file:{}/../../../cli/npm/envio", exe_path.to_string_lossy())
```
DZakh, line 241: By default it's installed using the `link:` protocol instead of file, but it makes the dependency resolution a little bit weird. The `file:` protocol makes pnpm treat the local envio as a real package.

Btw, if we add it for the `./generated` we won't need to have a `pnpm-workspaces.yaml` file.

codegenerator/integration_tests/tests/test_indexers/test_exits/pnpm-lock.yaml
```diff
@@ -1515,6 +1519,10 @@ snapshots:
     dependencies:
       once: 1.4.0
 
+  envio@file:../../../../cli/npm/envio:
+    dependencies:
+      rescript: 11.1.3
+
```
DZakh, line 1525: For comparison, `link:` didn't the rescript dependency in the pnpm-lock. I think this is what caused the issue with not being able to find the rescript dependency.

codegenerator/cli/src/executor/init.rs
JonoPrest: Btw in path_utils there's a function called normalize_path which will get rid of unneeded ../../ etc in the middle of your path. Doesn't matter too much but it's there.

JonoPrest: Awesome, never new about the "file:" protocol.

codegenerator/cli/npm/envio/local-bin.mjs
```diff
@@ -4,21 +4,69 @@
...
+    envioVersion = pnpmList[0].dependencies["envio"].version;
+  } catch (e) {
+    throw new Error(
+      `Failed to get envio version from pnpm list envio --json output. Error: ${e.message}. Pnpm list: ${pnpmList}`
+    );
+  }
+
+  if (!envioVersion.startsWith("file:")) {
+    throw new Error(
+      `Unexpected envio version. It should have the file: protocol. Actual: ${envioVersion}`
```
JonoPrest, line 54: Perhaps this can contain a link or just an example? I would be a little confused getting this error. Some people might already have run the init command and have an existing project. I have a few for eg.

codegenerator/cli/npm/envio/package.json.tmpl
JonoPrest: Is this still needed after adding file protocol?

codegenerator/cli/npm/envio/local-bin.mjs
DZakh: Honestly I didn't thought about the error handling deeply, left it to Supermaven generate most of them. I think the error is fine, since if you have the `"example"` version, then you won't run the binary in any way. If you have a simple path, the you'll get the `It should have the file: protocol`.

PR #279, opened 2024-10-23
Fix Contract Import to stop asking for API keys

DZakh:
Also:
- Changed the order. Contract Import is now first in the prompt, Local import is the second
- Don't abort the contract import flow on error, instead fallback to local import
- Added contract import support for:
  - ArbitrumTestnet
  - Boba
  - BscTestnet
  - CeloAlfajores
  - FantomTestnet
  - LineaSepolia
  - Mbase
  - Metis
  - Moonriver
  - PolygonZkevmTestnet + Fixed the chain id
  - ScrollSepolia
  - '+' 16 chains with blockscout explorer


DZakh:
Not ready yet. Forgot to change localhost to envio.dev. I also think to use an env variable just in case

DZakh:
Done

codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
```diff
@@ -74,38 +78,64 @@ impl ContractImportArgs {
...
-    ) -> Result<SelectedContract> {
-        match SelectedContract::from_etherscan(network_with_explorer, contract_address.clone())
-            .await
-        {
+    ) -> anyhow::Result<SelectedContract> {
+        let result = match contract_import(network, &contract_address, 0).await {
+            Ok(ContractImportResult::Contract(contract_data)) => Ok(contract_data),
+            Ok(ContractImportResult::NotVerified) => {
+                println!("Failed to find the verified contract on a block explorer.");
+                Err(())
```
JonoPrest, line 90: Can you rather return the string in the Err() and just use one `println!` before the next prompt ?

Just my preference to limit side effects where possible and it only made sense to me why you were printing a couple of lines down from here.

codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
```diff
@@ -74,38 +78,64 @@ impl ContractImportArgs {
...
+                .await;
+            }
+        };
+
+        let mut abi_events: Vec<ethers::abi::Event> = contract_data.abi.events().cloned().collect();
+        if !self.all_events {
+            abi_events = prompt_abi_events_selection(abi_events)?;
         }
+
+        let network_kind = get_converter_network_u64(network.clone() as u64, &None, &None)?;
```
JonoPrest, line 125: I miss rescript labelled arguments when reading function calls like this 🥲

JonoPrest (approved):
Awesome stuff! Looks great Dmitry, really looking forward to having this in.

PR #280, opened 2024-10-23
Improve crash error

DZakh:


moose-code (approved):
Amazing, LGTM

JonoPrest:
@DZakh, looks good but this is on a success exit it seems. So like when the indexer reaches endblock for eg.

PR #292, opened 2024-10-25
feat: set unordered head mode to true by default

DenhamPreen:
sorry took me some time to get to this,

I string searched the rollback setting and ammended it where ever I saw that then did a sanity check to see if I might have missed it elsewhere. I'm confident it should be set everywhere but maybe you're aware of somewhere I'm not thinking of :)


JasoonS (approved):

DZakh (commented):
There's a known potential issue with unordered multichain mode and rollback on reorg enabled together, where it might double-process or skip an event on rollback.

JonoPrest:
@DenhamPreen, as Dmitry says there's a known potential issue with rollbacks in unordered multichain mode. Can we wait until it's patched before including this in a release?

DenhamPreen:
Sure.
@JonoPrest, could you message me when I can update the branch :)

Issue #283, opened 2024-10-23
Bad rescript error message 


moose-code:
This error "Failed to seralize JSON at root. Reason: the S.transform serliazer is missing" 

has confused two different users now and doesn't provide a clear root as to the issue. 

This occured on the velodrome indexer and the rhinestone indexer. I believe @DZakh has an idea for why it occurs and a fix. 

![image](https://github.com/user-attachments/assets/75f5d674-ac7b-4d1b-8feb-6c0ced951f7b)

moose-code:
https://github.com/rhinestonewtf/rhinestone-indexer/commit/3aac2f6b7ebd9f5bcb4360e3e856295c90dd0f9a

Happened also when things weren't correctly awaited

JonoPrest:
Closed by https://github.com/enviodev/hyperindex/pull/297

PR #297, opened 2024-10-25
Improve Batch Set error lil bit

DZakh:
It became:
image

Instead of:
image

Still not perfect, but much better than before.

I think after the rescript-schema v9 release, we'll be able to improve it even more. It'll allow run serializing with type validation, so if we catch this null constraint error from pg, we can run the serializing with type validation and get the index of the entity where it fails. If we include it in the error, it should definitely help with debugging. Unfortunately adding the event id is difficult here.


DZakh:
Fixes https://github.com/enviodev/hyperindex/issues/283

JonoPrest (approved):
Awesome thank you! Much better error for sure 👍🏼

Issue #221, opened 2024-09-22
Composite indexes do not retain order


JasoonS:
This graphql schema

```grapgql
type TypeX @index(fields: ["a", "b"]) {
    id: ID!
    a: String!
    b: String!
}

type TypeY @index(fields: ["b", "a"]) {
    id: ID!
    a: String!
    b: String!
}
```
leads to the same indexes (a, b) for both tables

```
SELECT indexdef FROM pg_indexes WHERE tablename = 'TypeX';

CREATE INDEX "TypeX_a_b" ON public."TypeX" USING btree (a, b)


SELECT indexdef FROM pg_indexes WHERE tablename = 'TypeY';

CREATE INDEX "TypeY_a_b" ON public."TypeY" USING btree (a, b)
```

The fields ordering when creating a composite index should remain constant.

This is important for postgres query optimization - and can have an impact on how efficiently filtering inquiries works.

PR #301, opened 2024-10-28
Fix sorting of composite index fields

JonoPrest:
Closes https://github.com/enviodev/hyperindex/issues/221


codegenerator/cli/src/config_parsing/entity_parsing.rs
```diff
@@ -860,7 +860,7 @@ pub struct MultiFieldIndex(Vec<String>);
 
 impl MultiFieldIndex {
     fn new(field_names: Vec<String>) -> Self {
-        Self(field_names.into_iter().sorted().collect())
```
JonoPrest, line 863: This shouldn't have been sorted.

DZakh (approved):

PR #302, opened 2024-10-28
Use double precision for float representation

### LINKED ISSUE #187: Float from indexer is returned as numeric field
Screenshot 2024-09-09 at 16 50 41
Screenshot 2024-09-09 at 16 51 05

JonoPrest:
Closes #187 

- We do now have BigDecimal type that uses numeric so if "Double Precision" is not enough they can just use that.
- Double Precision maps to a  "Float" scalar in Hasura so it solves the given problem.
- There is already a test in test_codegen on the entity `EntityWithAllTypes` that validates de/serialising and writing to the db.
- Double Precision is also handled by the `HASURA_GRAPHQL_STRINGIFY_NUMERIC_TYPES` variable we already set for hasura.

This should probably at least go into a minor release since it could break an indexer that already uses the "Float" type.


DZakh (approved):
Nice ⚡

PR #308, opened 2024-10-30
Sync hypersync chain list +14 new chains

DZakh:
+ Add a test to not forget about this in the future. Might be annoying, but better than nothing


codegenerator/cli/src/config_parsing/chain_helpers.rs
```diff
@@ -90,7 +93,6 @@ pub enum Network {
     #[subenum(HypersyncNetwork)]
     C1Milkomeda = 2001,
 
-    #[subenum(NetworkWithExplorer)]
     Canto = 7700,
```
DZakh, line 96: Their explorer doesn't pass my tests

JonoPrest (approved):
Great, thanks for the extra safety as well 💪🏼

PR #314, opened 2024-11-01
Update chains and move test to integration tests

JonoPrest:


codegenerator/integration_tests/src/main.rs
```diff
@@ -137,8 +137,23 @@ async fn run_all_init_combinations() {
     }
 }
 
+async fn all_hypersync_networks_are_included() {
+    let diff = envio::scripts::print_missing_networks::get_diff()
+        .await
+        .unwrap();
+
+    if diff.missing_chains.is_empty() && diff.extra_chains.is_empty() {
+        return;
```
JonoPrest, line 146: Also just a note I fixed this to use `&&` instead of `||` since it was not failing when it should have.

DZakh (approved):

PR #332, opened 2024-11-11
Fix integration tests

JonoPrest:


DZakh (approved):
Is it because of the Address? Maybe we can remove `generated` now? Or there are some modules which still might be used?

JonoPrest:
> Is it because of the Address? Maybe we can remove `generated` from dependencies now? Or there are some modules which still might be used?

Still Types and Handlers are definitely needed from generated, so probably it will always be needed.

DZakh:
True

PR #334, opened 2024-11-12
Update fuel-ts up to 0.96.1

DZakh:
This might solve the Bako's issue


JonoPrest (approved):
Sweet, I assume no updates needed to the vendored fuel decoder?

PR #335, opened 2024-11-12
Mark MevCommit as HyperSync

DZakh:
I don't know why it disappears and appears from the list, but it should fix the integration tests. Maybe we should adjust the tests only to complain when we have missing non-experimental chains?


JonoPrest (approved):
Ah interesting, yeah I removed it also due to integration test issue

PR #339, opened 2024-11-15
Remove debug loggers

DZakh:
Noticed this in the logs:
image

It is also worth investigating why it's called so many times.


DZakh:
This is from http://localhost:3000/app/props-labs/props-mira-swaps/cf3dd6f/runtime-logs

JonoPrest (approved):
Yeah, probably means we're making more get height queries than necessary.

Issue #341, opened 2024-11-18
typescript type import error on `./bindings/OpaqueTypes.ts`


JasoonS:
**Describe the bug**
When running `pnpm tsc` on an indexer with v2.7.4 I get this:
```
[10:36:53] Starting compilation in watch mode...

node_modules/.pnpm/envio@2.7.4_typescript@5.2.2/node_modules/envio/src/Address.gen.ts:6:35 - error TS2307: Cannot find module './bindings/OpaqueTypes.ts' or its corresponding type declarations.

6 import type {Address as $$t} from './bindings/OpaqueTypes.ts';
```

Luckily it is benign, and doesn't affect runtime.

DZakh:
This is because we include only `.res` files to the package. Probably should include the whole src dir.

<img width="337" alt="image" src="https://github.com/user-attachments/assets/6ac9edb6-16d2-4b47-b3b4-d4b0296f1e11">

PR #344, opened 2024-11-18
Fix npm files to include genType

DZakh:
Solves https://github.com/enviodev/hyperindex/issues/341


codegenerator/cli/npm/envio/package.json.tmpl
```diff
@@ -39,6 +39,8 @@
     "evm.schema.json",
     "fuel.schema.json",
     "rescript.json",
-    "src/**/*.res"
+    "src/**/*.res",
+    "src/**/*.resi",
+    "src/**/*.ts"
```
DZakh, line 44: Maybe this would work:
```
src
!src/**/*.bs.js
```
But I'm not sure, so I kept it simple

JonoPrest: Hmmm, are there some files that we are trying to omit from this package? Maybe it's better to just exclude those. For eg, moving a plain js file in with a binding to it (like db functions for eg) won't work and we need to remember to do this otherwise we only discover it on publish.

JonoPrest (approved):
Nice, thanks for fixing this 👍🏼

codegenerator/cli/npm/envio/package.json.tmpl
DZakh: Yeah, we can add `.npmignore` later. I think we only want to exclude a few files and tests in the future.

PR #347, opened 2024-11-19
More flexible package files

DZakh:
I've started with .npmignore, but it had some gotchas https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

So I've done it another way. Tested it locally, works correctly


JonoPrest (approved):
Sweet, if this works then great 👍🏼

Issue #315, opened 2024-11-01
Feature Request: Templating engine  / support for environment variables via .env files


Simon0x:
Hello, 

this is not a top priority but after talking to multiple developers of what envio is capable of doing, I noticed that a lot would like to have a templating engine and/or support for environment variables.

I think this could be a great feature to adopt usage of Envio compared to subgraphs.
I've setup a workaround in my project which I can share here, however I think it would be great to have an official solution.

<details>
  <summary>Adujustments to package.json</summary>

```
"codegen": "pnpm run config && envio codegen",
"dev": "npx dotenvx run -- envio dev",
"start": "npx dotenvx run -- ts-node generated/src/Index.bs.js",
"config": "npx dotenvx run -- ts-node ./scripts/render-config-template.ts"
```
</details>

<details>
  <summary>.env file</summary>

```
# This is either mainnet or testnet
TEMPLATE=mainnet
```
</details>

<details>
  <summary>render-config-template.ts</summary>

```
import fs from 'fs';
import ejs from 'ejs';
import { Config } from '../src/types/config.ts';
import { getLoadedConfig } from '../src/config';

type ConfigOverride = Config & {
  jointUniswapPoolContracts: string;
};

if (!process.env.TEMPLATE) {
  console.error('No TEMPLATE environment variable set');
  process.exit(1);
}

if (process.env.TEMPLATE !== 'mainnet' && process.env.TEMPLATE !== 'testnet') {
  console.error('Invalid TEMPLATE environment variable set');
  process.exit(1);
}

const loadedConfig = getLoadedConfig();

const configValues: ConfigOverride = {
  ...loadedConfig,
  jointUniswapPoolContracts: [loadedConfig.uniswapWethBstPoolAddress, loadedConfig.uniswapBstUsdtPoolAddress].join(','),
};

const configTemplate = fs.readFileSync(
  `./templates/template.${process.env.TEMPLATE}.config.yaml`,
  'utf8'
);

const configString = ejs.render(configTemplate, { config: configValues });

const generationWarning = `### This file is autogenerated from the template.config.yaml file in the root of the project ###
### Edit the template.config.yaml file in the root of the project and run \`pnpm run codegen\` to update this file ###
`;

try {
  fs.writeFileSync('./config.yaml', generationWarning + configString);
} catch (e) {
  console.log(e);
}

```
</details>

<details>
  <summary>template.mainnet.config.yaml</summary>

```
name: Blocksquare
field_selection:
[...]

PR #350, opened 2024-11-20
Interpolate env vars into config

DZakh:
Now, it's possible to add `${MY_ENV_VAR}` in any place of the config file, and it'll be replaced with an environment variable on codegen.

1. First the variable will be looked up in the existing environment
2. If the variable is not present, it'll attempt to find it in the `.env` file at the root of the project
3. On failure it'll list all the missing env variables

Multiline envs should be supported as well. So maybe it's even possible to put the whole config to the .env file 😅

Some things to consider:
1. IDE will show validation errors on `${}` usage in some places (scoped out)
2. I don't really like that we load the `.env` file even when the `${}` is not used in the config. Should I fix this? (solved)
3. I Just realised that we should use the interpolated version for the hash, otherwise the indexer won't recodegen on the .env value changes. I'll fix it now. (solved)


codegenerator/cli/Cargo.toml
```diff
@@ -8,7 +8,6 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.2.2", features = ["derive", "env"] }
 clap-markdown = { version = "0.1.0" }
-dialoguer = "0.8"
```
DZakh, line 11: Wasn't used

codegenerator/cli/Cargo.toml
```diff
@@ -44,9 +43,7 @@ thiserror = "1.0.50"
 fuel-abi-types = "0.7.0"
 schemars = { version = "1.0.0-alpha.2", features = ["preserve_order"] }
 convert_case = "0.6.0"
-# TODO: replace ethers with alloy and foundry-block-explorers since these are actively maintained
-# foundry-block-explorers = "0.5.1"
-# alloy-chains = "0.1.23"
```
DZakh, line 49: Not relevant anymore

DZakh:
Fixes https://github.com/enviodev/hyperindex/issues/315

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -447,6 +449,46 @@ impl SystemConfig {
...
+                .context("Failed loading environment variables")?,
+            err => err.context("Failed loading .env file")?,
+        };
+        Ok(env_map)
+    }
+
+    fn interpolation_config_variables(config_string: String, env_map: EnvMap) -> Result<String> {
+        let mut missing_vars = Vec::new();
+
+        let re = Regex::new(r"\$\{([a-zA-Z_][0-9a-zA-Z_]*)\}").unwrap();
```
JonoPrest, line 471: TBH, I don't love regex. Because I never fully understand it 😓 much prefer parser combinator approach but I do understand how useful regex is.

If regex is used I'd prefer it to be moved into it's own function with unit tests specifically for that function.

Questions I have that tests would solve:
1. What are the boundaries of the match? Looks like for instance that inside the braces it won't accept symbols other than `_`? That I'm even not sure of... so would be good to show acceptable/unacceptable values in a test.
2. What if it doesn't match the inner braces but the user has added ${}? Looks like we don't have a good error to help the user with that.
3. How does it handle escaping? Not sure where but maybe I want to escape `\${` in my config. etc.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -447,6 +449,46 @@ impl SystemConfig {
...
+        let config_string = re.replace_all(&config_string, |caps: &Captures| {
+            match env_map.var(&caps[1]) {
+                Ok(val) => val,
+                Err(_) => {
+                    missing_vars.push(caps[1].to_string());
+                    (&caps[0]).to_string()
```
JonoPrest, line 477: What is the difference here between using cas[1] and caps[0]?

JonoPrest (approved):
Thanks Dmitry, looks good to me given the fix to hashing the interpolated config. The only real addition I would like is that there is a helpful error if `${}` is used but the inner value will not match the regex. 

> IDE will show validation errors on ${} usage in some places

Would be nice if we can find a composable way to allow interpolation for fields where IDE can complain. But I think the primary usecase is RPC endpoint which shouldn't complain. 

> I don't really like that we load the .env file even when the ${} is not used in the config. Should I fix this?

If you can think of a quick fix great but it's not a big deal.

> I Just realised that we should use the interpolated version for the hash, otherwise the indexer won't recodegen on the .env value changes. I'll fix it now.

👍🏼

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: caps[0] is the original content, caps[1] is the content of the first regex group (what's inside of the paranthes)

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: 1. `[]` is something like or in the RegExp world, so it'll either a-z, A-Z or _
2. Yes, indeed
3. This is currently not supported. We can add support for it later when somebody requests

DZakh: I personally used a parser combinator just a several times and find Regex much simpler. But this is probably because my first ever Js app heavely used regex to parse complex HTML 😅

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+            (Some(pos), _) => {
+                let name = &inner[..pos];
+                let default_value = inner[pos + 2..].to_string();
+                (
+                    name,
+                    InterpolationResult::DefaultForMissingAndEmpty(default_value),
+                )
+            }
+            (None, None) => (inner, InterpolationResult::DirectSubstitution),
+        };
```
DZakh, line 137: A little bit of code duplication, but I've decided to keep it as it is.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+DefaultForMissing with missing env and many dashes: "${MISSING_ENV----:---}"
+DefaultForMissing with missing env and empty default: "${MISSING_ENV-}"
+DefaultForMissingAndEmpty with existing env: "${EXISTING_ENV:-default}"
+DefaultForMissingAndEmpty with existing env and many dashes: "${EXISTING_ENV:----:---}"
+DefaultForMissingAndEmpty with missing env: "${MISSING_ENV:-default}"
+DefaultForMissingAndEmpty with missing env and many dashes: "${MISSING_ENV:----:---}"
+DefaultForMissingAndEmpty with missing env and empty default: "${MISSING_ENV:-}"
+DefaultForMissingAndEmpty with empty env: "${EMPTY_ENV:-default}"
+DefaultForMissingAndEmpty with empty env and many dashes: "${EMPTY_ENV:----:---}"
+DefaultForMissingAndEmpty with empty env and empty default: "${EMPTY_ENV:-}"
```
DZakh, line 336: Now we support the following options of vars interpolation:
image

The whole list is here: https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+                        {
+                            Ok(env_map) => {
+                                self.maybe_dotenv = Some(env_map.clone());
+                                env_map.var(name)
+                            }
+                            Err(err) => {
+                                match err {
+                              dotenvy::Error::Io(_, _) => (),
+                              _ => println!("Warning: Failed loading .env file with unexpected error: {err}"),
+                          };
```
JonoPrest, line 85: Think the indentation is off here, maybe running `make format` will help.

JonoPrest: So the idea here is that if it fails to load the file because it's not there that is fine but otherwise the error gets reported?

Just checking the warning won't be printed in this case.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+                                self.maybe_dotenv = Some(EnvMap::new());
+                                Err(err)
+                            }
+                        }
+                    }
+                };
+                match result {
+                    Ok(val) => Some(val),
+                    Err(_) => None,
+                }
```
JonoPrest, line 95: So nothing is done with the error case? What about propagating the errors as they go?

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+        let re = Regex::new(r"\$\{([^}]*)\}").unwrap();
+        let config_string = re.replace_all(&config_string, |caps: &Captures| {
+            let name = &caps[1];
+            let (name, interpolation_result) = parse_capture(name);
+            if interpolation_result == InterpolationResult::InvalidName {
+                // Wrap invalid vars with quotes to make them more visible in the error message
+                // Don't need to do this for missing ones, because they won't have spaces in the name
+                invalid_vars.push(format!("\"{name}\""));
+                return "".to_string();
+            }
```
JonoPrest, line 169: Should this maybe be put in the match arm?

Also what happens with `DirectSubstitute`? Maybe you could add the last cases for an exhaustiveness enforcement on the match instead of default?

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+                (None, _) => {
+                    missing_vars.push(name.to_string());
+                    "".to_string()
+                }
+            }
+        });
+
+        if !invalid_vars.is_empty() {
+            return Err(anyhow!(
+              "Failed to interpolate variables into your config file. Invalid environment variables are preset: {}",
```
JonoPrest, line 188:
```suggestion
              "Failed to interpolate variables into your config file. Invalid environment variables are present: {}",
```

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+              invalid_vars.join(", ")
+          ));
+        }
+
+        if !missing_vars.is_empty() {
+            return Err(anyhow!(
+              "Failed to interpolate variables into your config file. Environment variables are not preset: {}",
```
JonoPrest, line 195:
```suggestion
              "Failed to interpolate variables into your config file. Environment variables are not present: {}",
```

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: Exactly

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: It's either not present or not a unicode. I've decided to change it to Option for simpler mocking in tests.

image

codegenerator/cli/src/config_parsing/system_config.rs
DZakh: I don't want to call `get_env` when the name is invalid.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -45,10 +47,327 @@ pub enum Ecosystem {
...
+                (None, _) => {
+                    missing_vars.push(name.to_string());
+                    "".to_string()
+                }
+            }
+        });
+
+        if !invalid_vars.is_empty() {
+            return Err(anyhow!(
+              "Failed to interpolate variables into your config file. Invalid environment variables are present: {}",
```
JonoPrest, line 188: Maybe we can explain what would be a valid name

JonoPrest (approved):
🚀

PR #353, opened 2024-11-21
Move pg to npm package and vendore promise bindings

DZakh:


JasoonS (approved):
Awesome - it looks good on reading through. Didn't analyse super deeply - but in the type system I trust :muscle:

PR #356, opened 2024-11-22
WIP: Custom RPC client

DZakh:
- Downgrade rescript-schema because there's a bug with reverse converting some unions. Not related to us, but I was afraid that it might happen for some RPC requests
- Vendor rescript-rest and use it for RPC.res
- Vendor rescript-express


JonoPrest (approved):
Great 🥇

PR #360, opened 2024-11-23
chore: make project paths public

drewstone:


JonoPrest:
Hey @drewstone, could you explain why this is needed?

PR #362, opened 2024-11-25
Implement tiers and checks for networks

JonoPrest:
- Adds chain tiers and icon to network selection 
Screenshot 2024-11-25 at 12 51 07

- Uses API to manage only public chains/tiers and check tests

### DIFF
--- a/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
+++ b/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
@@ -25,7 +25,8 @@ use crate::{
 };
 use anyhow::{Context, Result};
 use inquire::{validator::Validation, CustomType, Select, Text};
-use std::{env, path::PathBuf, str::FromStr};
+use serde::{Deserialize, Serialize};
+use std::{env, fmt, path::PathBuf};
 use strum::IntoEnumIterator;
 
 fn prompt_abi_events_selection(events: Vec<ethers::abi::Event>) -> Result<Vec<ethers::abi::Event>> {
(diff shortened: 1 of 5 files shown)


codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
```diff
@@ -306,16 +316,7 @@ impl ExplorerImportArgs {
...
             None => {
-                let options = NetworkWithExplorer::iter()
-                    //Filter only our supported networks
-                    .filter(|&n| {
-                        HypersyncNetwork::iter()
-                            //able to cast as u64 because networks enum
-                            //uses repr(u64) attribute
-                            .find(|&sn| n as u64 == sn as u64)
-                            .is_some()
-                    })
```
DZakh, line 317: Hmm, I thought I deleted the code. Looks like I missed the place 👍

codegenerator/cli/src/config_parsing/chain_helpers.rs
```diff
@@ -457,11 +459,96 @@ impl Network {
...
+            | Opbnb | Lisk | BlastSepolia | Celo | Chiliz | Fuji | ArbitrumOne | Merlin
+            | Holesky => Silver,
+
+            Zora | MoonbaseAlpha | Morph | LuksoTestnet | Kroma | GnosisChiado | XLayer | Lukso
+            | Gnosis | C1Milkomeda | Crab | Tangle | Sophon | Flare | PolygonZkevm | MevCommit => {
+                Bronze
+            }
+
+            SophonTestnet | MorphTestnet | GaladrielDevnet | CitreaTestnet | Goerli
+            | BscTestnet | B2Testnet | UnichainSepolia => Experimental,
```
DZakh, line 516: Should we maybe store the json file and compute the values from it in runtime?

I think it's going to be annoying to sync the list. Also, we can do the same for whether chain is hypersync or not + add a make script to update the json file.

DZakh (approved):

codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
```diff
@@ -306,16 +316,7 @@ impl ExplorerImportArgs {
...
-                    .collect();
```
JonoPrest, line 318: I removed this filtering of only offering hypersync networks for contract import. It should now be obvious if they are hyper sync networks or not from the icon.

codegenerator/cli/src/config_parsing/chain_helpers.rs
JonoPrest: Would like to add things like the default confirmed block threshold to the API

DZakh: Yeah, that would be great. But inline it to binary, instead of fetching from the server on every run

DZakh: But we also have non-hypersync chains there, so I'm not really sure

PR #366, opened 2024-11-28
Fix integration tests

DZakh:


codegenerator/integration_tests/src/main.rs
```diff
@@ -153,8 +153,8 @@ async fn all_hypersync_networks_are_included() {
 
 #[tokio::main]
 async fn main() {
-    all_hypersync_networks_are_included().await;
     run_all_init_combinations().await;
+    all_hypersync_networks_are_included().await;
```
DZakh, line 157: Changed the order, so we can at least know that the `run_all_init_combinations` passes when `all_hypersync_networks_are_included` fails.

JonoPrest: I was actually thinking we should rather put these as separate tests in a suite and then run the tests.

DZakh: Yeah, I'll do it next time it annoys me 😅

JasoonS (approved):
Cool - yeah, can be annoying since hypersync chains can and will keep evolving. Still think a non-hardcoded list of chains is the way to go when we have time to implement

PR #367, opened 2024-11-28
Fix init messages call

DZakh:
Broke the request a little bit


DZakh:
> When did this change?

When I started using rescript-rpc in the previous release. Currently broken in latest

DZakh:
I'll add tests for the call

PR #411, opened 2025-01-09
Fix indexer stopping fetching nested dynamic contract registrations on restart when a preRegistration is used

DZakh:
Bug reproduction:
1. Have an indexer with a preRegistering dynamic contract
2. The preRegistering one should register a contract which registers another contract
3. If we stop the indexer and start it again, the nested registered contract won't be indexing

Terminology:
- dc - dynamic contract

The issue was caused by incorrect dc recovering logic when we have an event with preRegistration.
Without preRegistration we recover all dcs registered before or equal to the start block, while with preRegistration event this logic is changed to get all events with preRegistration and extract them from the db. Even though it sounds broken, but this worked in 95% of the usecases, besides when we got new dc registrations after preRegistration. It caused two bugs:
1. If a dc contract is registered after preRegistration and NONE of its events have preRegistration enabled, it won't be loaded from db on the restart recovery
2. If a dc contract is registered after preRegistration and SOME of its events have preRegistration enabled, in some cases it will be loaded from db even after the restart start block

To solve the problem:
1. I've added the `is_pre_registered` flag to the dc entity in the db, so it's possible to correctly get all preRegistered dcs from db correctly for both 1 and 2 cases mentioned above.
2. Fixed overwriting the dc in db when it's already registered. Wasn't really an issue before, besides wasted resources, but now it caused overwriting the `is_pre_registered` flag

Unrelated issue I solved:
If a single event registers multiple dcs with the same addresses, the duplicate wasn't detected correctly.

### DIFF
--- a/codegenerator/cli/templates/dynamic/codegen/src/ContextEnv.res.hbs
+++ b/codegenerator/cli/templates/dynamic/codegen/src/ContextEnv.res.hbs
@@ -6,7 +6,6 @@ The context holds all the state for a given events loader and handler.
 type t = {
   logger: Pino.t,
   eventItem: Internal.eventItem,
-  addedDynamicContractRegistrations: array<TablesStatic.DynamicContractRegistry.t>,
 }
 
 let getUserLogger = (logger): Logs.userLogger => {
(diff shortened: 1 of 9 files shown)


JasoonS (approved):
:pray:

PR #427, opened 2025-01-17
Add support for RPC input & value transaction fields

DZakh:
It happened to be already supported, I don't know why I didn't allow them in the beginning 😅


JonoPrest (approved):
Awesome  💪🏼

PR #432, opened 2025-01-21
Rename chain worker to source and make it a record

DZakh:
To better reflect what it is and make it more convenient to use.


JonoPrest (approved):
Looks great thanks!

PR #456, opened 2025-02-13
Fix events with empty params

DZakh:
rescript-schema v9 started to have stricter validation for literal fields, which made it fail when trying to convert empty event params to raw events. The failure was because of another really old bug in the `convertHyperSyncEventArgs` where we simply passed through `decodedEvent` instead of returning `()`.


DenhamPreen (approved):
🙌

PR #458, opened 2025-02-13
Enable rollback on reorg in init templates

DZakh:


DenhamPreen (approved):
Should we do it for contract import then too?

DZakh:
> Should we do it for contract import then too?

Done

PR #461, opened 2025-02-28
Clean up main README.md

DZakh:
This is now a symlink to the `envio` package README page.

https://github.com/enviodev/hyperindex/tree/dz/clean-up-landing-readme

I moved everything else to the `CONTRIBUTING.md` and did a little cleanup. This should be cleaned up even more, but it needs some additional special efforts.


CONTRIBUTING.md
```diff
@@ -0,0 +1,112 @@
...
+
+   ```sh
+   npm install --global pnpm
+   ```
+
+3. Cargo https://doc.rust-lang.org/cargo/getting-started/installation.html
+
+   ```sh
+   curl https://sh.rustup.rs -sSf | sh
+   ```
```
DenhamPreen, line 17: nit: I think it would make sense to have this as step 4 possibly since the previous 2 & docker are more related to running the indexer

CONTRIBUTING.md
```diff
@@ -0,0 +1,112 @@
...
+
+To update all generated docs run
+
+`make update-generated-docs`
+
+To updated just the config json schemas
+`make update-schemas`
+
+Or to update just the cli help md file
+`update-help`
```
DenhamPreen, line 56:
```suggestion
`make update-help`
```
?

DenhamPreen (approved):
Nice!

CONTRIBUTING.md
DZakh: We should revisit the file since it's very outdated

PR #463, opened 2025-03-04
House keeping: Fix all clippy warnings

JonoPrest:
Ran the clippy linter for the codegen code and fixed all its warnings.

### DIFF
--- a/codegenerator/cli/src/config_parsing/system_config.rs
+++ b/codegenerator/cli/src/config_parsing/system_config.rs
@@ -140,7 +140,7 @@ mod interpolation {
         };
 
         if name.is_empty()
-            || name.chars().next().map_or(false, |c| c.is_ascii_digit())
+            || name.chars().next().is_some_and(|c| c.is_ascii_digit())
             || !name.chars().all(|c| match c {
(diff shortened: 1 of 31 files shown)


codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -1118,7 +1118,7 @@ impl EvmAbi {
 #[derive(Debug, Clone, PartialEq)]
 pub enum Abi {
     Evm(EvmAbi),
-    Fuel(FuelAbi),
+    Fuel(Box<FuelAbi>),
```
DZakh, line 1121: I never used `Box`. What is it needed for?

DZakh (approved):
Wow, this is impressive 🔥

codegenerator/cli/src/config_parsing/system_config.rs
JonoPrest: It's a clippy warning suggestion.

Box is like a c pointer. So it's an address on the heap. The reason to use it here is that the FuelAbi type is very big in memory. So the enum has to allocate all that memory even in the evm case. Clippy suggests using a pointer here instead to be more efficient

PR #466, opened 2025-03-06
Feat: Add support for Foundry output ABI format

diyahir:
This PR adds support for Foundry output ABI format, which places the ABI definition in an "abi" child field within the JSON file. Previously, the code would fail early when trying to parse these files, as it expected the ABI to be at the root level of the JSON.

## Changes

- Enhanced the ABI parsing logic in `LocalImportArgs::parse_contract_abi` to handle nested ABI structures
- Added a two-step parsing approach:
  1. First attempt to parse the JSON directly as an ABI
  2. If that fails, try to parse it as a JSON object that might contain an "abi" field
- Added comprehensive error messages to help users troubleshoot ABI parsing issues
- Added test cases to verify both direct and nested ABI parsing works correctly

## Test Coverage

Added two test cases to verify the functionality:
- `test_parse_contract_abi_direct`: Tests parsing a standard ABI file
- `test_parse_contract_abi_nested`: Tests parsing an ABI file with the "abi" field at the root level (Foundry format)

Also added test fixtures:
- Added a sample Foundry-style ABI file (`Contract3.json`) with the ABI in an "abi" child field
- Added a corresponding test configuration (`nested-abi.yaml`) to test the system config parsing

### DIFF
--- a/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
+++ b/codegenerator/cli/src/cli_args/interactive_init/evm_prompts.rs
@@ -333,47 +333,36 @@ impl ExplorerImportArgs {
 
 impl LocalImportArgs {
     fn parse_contract_abi(abi_path: PathBuf) -> anyhow::Result<ethers::abi::Contract> {
-        let abi_file = std::fs::read_to_string(&abi_path).context(format!(
+        use crate::config_parsing::system_config::EvmAbi;
+        use crate::project_paths::ParsedProjectPaths;
+
+        let _abi_file = std::fs::read_to_string(&abi_path).context(format!(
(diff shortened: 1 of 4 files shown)


DZakh:
Good morning. Thank you for your contribution! Really appreciated 🙏
A single comment and request is that we already have the logic for parsing ABI already set in the config. It would be nice to unify it with contract import you added, so the logic is not duplicated.

Here's the existing code: https://github.com/enviodev/hyperindex/blob/2e52973267e47aec362c22b0a84dde3ed1c812c1/codegenerator/cli/src/config_parsing/system_config.rs#L1075-L1114

diyahir:
Updated

DZakh:
@diyahir, thank you for your contribution. I reimplemented your changes in another PR to fit it into the next release.

PR #489, opened 2025-03-20
Usel lodash temporarily for deep cloning

JonoPrest:
This fixes a bug where `BigDecimal` objects had their prototypes stripped during nodejs "structuredClone".

It had the `isEqualTo` method stripped away which caused a runtime failure on the inMemoryTable indexes during a test.

@DZakh I've added lodash as a dep to use their `cloneDeep` method but I think we could use rescript schema to generated deep clone functions.


JonoPrest:
Note I've removed it as a dep and vendored just the function we need. (Still over a thousand lines of js we could maybe ✂️ )

DenhamPreen (approved):

PR #504, opened 2025-04-11
Support passing additional params for logger calls

DZakh:
Moved the `Logger` type to the `envio` package.

```ts
import type { Logger } from "envio"
```

Added the ability to pass params on a log call:

```ts
context.log.info("Sucessfully handled Transfer()", { from: event.params.from, to: event.params.to })
```

These params will be displayed in the logs in your terminal as well as Hosted Service.

You can also pass an error:

```ts
} catch (error) {
  context.log.error("Failed ipfs call", error)
}
```

Removed `context.log.errorWithExn` from Js/Ts API. Removed the `Logs` module from ReScript API and introduced some breaking changes to the arguments' order.


codegenerator/cli/npm/envio/src/Types.ts
```diff
@@ -0,0 +1,22 @@
+export type Invalid = never;
+
+export type Address = string;
```
JonoPrest, line 3: I was wondering if at we should use the same type as viem for ts address? Like `0x${string}`

Might be breaking though. But users sometimes have to type cast to turn viem address into string for eg.

JonoPrest (approved):
Looks nice! 🙏🏼

codegenerator/cli/npm/envio/src/Types.ts
DZakh: A good thing to consider. I don't think it's going to be a breaking breaking change

PR #507, opened 2025-04-22
IDX-830 Add Prometheus metrics

DZakh:
- `envio_indexing_addresses` - The number of addresses indexed on chain. Includes both static and dynamic addresses.
- `envio_indexing_end_block` - The block number to stop indexing at. (inclusive)
- `envio_progress_block_number` - The block number to track the progress of indexing at. Currently uses the fully fetched block number. In the future will be changed to block number processed and stored in the database.

Also, stopped closing completed indexer process when used in dev mode with TUI


JasoonS:
Cool, it makes sense. I went through the issue and edited things. I think this first part can be merged though.

Am I understanding correctly `envio_progress_block_number` - Currently is the latest block fetched from the chain or the source. But in the future you want this to be the latest block number that is actually processed by the handlers and finished saving to the database?

JasoonS:
I updated the issue, let me know if you agree: https://linear.app/envio/issue/IDX-830/update-hyperindex-code-with-new-metrics

JasoonS (approved):

DZakh:
> Cool, it makes sense. I went through the issue and edited things. I think this first part can be merged though.
> 
> Am I understanding correctly `envio_progress_block_number` - Currently is the latest block fetched from the chain or the source. But in the future you want this to be the latest block number that is actually processed by the handlers and finished saving to the database?

Exactly

DZakh:
> I updated the issue, let me know if you agree: https://linear.app/envio/issue/IDX-830/update-hyperindex-code-with-new-metrics

Fixed some things.

PR #509, opened 2025-04-25
Fix field selection with RPC source

DZakh:
Fix the error and improve the error message to be visible

image

Original:

image


JasoonS:
OK makes sense to remove that conditional to show the source error.

PR #519, opened 2025-05-09
Add a trace log if a polling request hangs

JonoPrest:
Thought this would be helpful since I have a case where hypersync was not lagging at all but the indexer was reporting that the source height did not increase in > 20s. And there were no failures on the polling.

I wanted a simple way to check if the polling was hanging for some time or if it was successful.


DZakh (approved):
I don't really like to create a timeout here, but definitely useful for debugging. Although, maybe we use prom metric with time measurement instead?

JonoPrest:
> I don't really like to create a timeout here, but definitely useful for debugging. Although, maybe we use prom metric with time measurement instead?

I don't mind how it's done, just need to be able to check for a given height request if the polling was hanging

DZakh:
I'll merge it like this and then change it to Prometheus metric if I have time

PR #520, opened 2025-05-12
IDX-861 - Move to using built-in fetch from node (rather than polyfill)

JasoonS:


JasoonS:
I need to update the docs, make this a 'hard' requirement (not just recommended that it is v18 or higher):
image

DZakh (approved):

PR #522, opened 2025-05-13
Add index.js to the published package template

DZakh:


JonoPrest (approved):
How was this working before? Is it new?

DZakh:
Yes, it's new with Effect API

DZakh:
Previously everything was imported from generated

PR #525, opened 2025-05-15
Add test to show the output of the generated query

JonoPrest:
I wanted to inspect these queries so I added a test


DZakh (approved):
Nice, sorry that I didn't add it in the beginning.

PR #540, opened 2025-05-20
Generate type for loaderArgs and handlerArgs per event

JonoPrest:
Quick one following up from this discussion: https://github.com/enviodev/hyperindex/discussions/539

When creating a loader outside of the handler registration, using the generated types means you can't use type inference for loader return.

Eg. from Paul: 

```ts
type LoaderReturn = {
  asset: Entity.Asset | undefined;
  batch: Entity.Batch | undefined;
  batcher: Entity.Batcher | undefined;
  watcher: Entity.Watcher | undefined;
};

const loader: SablierLockup_v2_0_CreateLockupTranchedStream_loader<LoaderReturn> = async ({
  context,
  event,
}): Promise<LoaderReturn> => {
  const assetId = Id.asset(event.srcAddress, event.chainId);
  const asset = await context.Asset.get(assetId);

  const sender = event.params.commonParams[1];
  const batchId = Id.batch(event, sender);
  const batch = await context.Batch.get(batchId);

  const batcherId = Id.batcher(event, sender);
  const batcher = await context.Batcher.get(batcherId);

  const watcherId = event.chainId.toString();
  const watcher = await context.Watcher.get(watcherId);
  return {
    asset,
    batch,
    batcher,
    watcher,
  };
};
```

It's more ergonomic to add a type to just the args and let TS infer the return.

I've added the type for handlerArgs as well (although not sure how useful it is because you would need to define loaderReturn)

## Summary by CodeRabbit

- **Refactor**
  - Improved the structure of type alias declarations for event loaders and handlers, making them more modular and easier to understand. No changes to existing functionality.


DZakh (approved):
I don't really like that they will be exposed in lowercase, but I think we'll be able to sort it out in v3.

DZakh:
I see that they are not exposed with the `preload_handlers: true`. I'll add the types back in the next release.

PR #552, opened 2025-05-26
Ensure history maintains same order as processing

JonoPrest:
I realised during testing, that within a batch history items can be inserted out of order of order (between different entities in the same table).

I could not find a place where this would matter to the rollback logic, but there are a lot of assumptions in various queries in the code that the serial of the entity history row matches the ordering of the events.

I honestly don't know if this will make any difference. There is another issue in that the dynamic contracts table can be out of  order (which is treated as an entity in a rollback.) I have yet to find if this has any link to rollback issues.


DZakh (approved):
Interesting. Let's try it out 👍

PR #554, opened 2025-05-27
Librarify the entity config type

DZakh:
## Summary by CodeRabbit

- **Refactor**
  - Unified entity handling by replacing module-based parameters with explicit entity configuration objects throughout the codebase.
  - Updated function signatures and types to use the new entity configuration approach for improved consistency and type safety.
  - Adjusted error messages and logging to reference entity configuration properties directly.
- **Tests**
  - Updated all related tests to use the new entity configuration pattern and internal entity representations.

### DIFF
--- a/codegenerator/cli/npm/envio/src/ReorgDetection.res
+++ b/codegenerator/cli/npm/envio/src/ReorgDetection.res
@@ -70,6 +70,7 @@ module LastBlockScannedHashes: {
     t,
     ~blockNumbersAndHashes: array<blockDataWithTimestamp>,
     ~currentBlockHeight: int,
+    ~skipReorgDuplicationCheck: bool=?,
   ) => validBlockResult
 
   let getThresholdBlockNumbers: (t, ~currentBlockHeight: int) => array<int>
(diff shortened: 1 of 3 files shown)


codegenerator/cli/templates/dynamic/codegen/src/db/Entities.res.hbs
```diff
@@ -11,9 +11,9 @@ module type Entity = {
   let table: Table.table
   let entityHistory: EntityHistory.t<t>
 }
-module type InternalEntity = Entity with type t = internalEntity
-external entityModToInternal: module(Entity with type t = 'a) => module(InternalEntity) = "%identity"
-external entityModsToInternal: array<module(Entity)> => array<module(InternalEntity)> = "%identity"
+external entityModToInternal: module(Entity with type t = 'a) => Internal.entityConfig = "%identity"
+external entityModsToInternal: array<module(Entity)> => array<Internal.entityConfig> = "%identity"
+external entitiesToInternal: array<'a> => array<Internal.entity> = "%identity"
```
JonoPrest, line 16: I'm a little worried about type casting modules to records. Not sure if there's any possibility of unexpected compiled differences between the two.

Do we need the module at all if we are changing to records or is it just about keeping naming conventions with uppercase module etc?

DZakh: It's just to keep the change smaller. Ultimately, we should also convert the `module Entity` into a record.

JonoPrest (commented):
Could you explain what benefit this refactor has? I understand why it can be nicer to work with records over first class modules but it looks like we still use the first class modules and then just type cast them which IMO seems to just reduce type safety.

JonoPrest (approved):
All good thanks 🙏🏼

PR #573, opened 2025-06-05
Remove dependency on root from generated

DZakh:
This is the first PR to solve the issue of using HyperIndex as a part of another monorepo - https://github.com/enviodev/docs/issues/650

This solves a cycle dependency between the root and the generated project, as well as reduces the reliance of the generated dir on pnpm.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration file to provide a dynamic relative path for generated code.

- **Refactor**
  - Updated path resolution for requiring handler files and locating the `.env` file to use the new dynamic relative path instead of hardcoded values.

- **Chores**
  - Removed an unused dependency entry from the package template.

### DIFF
--- a/codegenerator/cli/templates/dynamic/codegen/package.json.hbs
+++ b/codegenerator/cli/templates/dynamic/codegen/package.json.hbs
@@ -55,7 +55,6 @@
     "rescript": "11.1.3",
     "rescript-envsafe": "5.0.0",
     "rescript-schema": "9.3.0",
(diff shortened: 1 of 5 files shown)


JonoPrest (approved):
Great! Will be glad to have a use case that can flush out these kinds of issues. The initial design I think had monorepos with multiple indexers in mind but some of these requirements fell by the way side with nobody using it like that.

PR #578, opened 2025-06-05
Make HyperIndex work without pnpm workspace

DZakh:
## Summary by CodeRabbit

- **New Features**
  - Added automatic inclusion of the current envio version in generated projects.
  - Projects now explicitly declare themselves as private and include envio as a dependency.

- **Improvements**
  - Enhanced version validation and retrieval for envio, ensuring accurate versioning in generated code.
  - Dynamic path handling for ABI imports in generated Fuel contracts.
  - Extraction now skips `.gitkeep` files to avoid unnecessary files in new projects.
  - Updated package installation process to improve dependency management during code generation.

- **Chores**
  - Removed multiple `.npmrc` and `pnpm-workspace.yaml` files from templates and scenarios, streamlining workspace and package manager configurations.
  - Cleaned up configuration files by removing unused or redundant fields.
  - Narrowed workspace scope in specific scenarios to focus on generated packages.


codegenerator/cli/src/commands.rs
```diff
@@ -114,15 +116,7 @@ pub mod codegen {
             return Ok(exit1);
         }
 
-        println!("Clean build directory");
-        let exit2 = rescript::clean(&project_paths.generated)
-            .await
-            .context("Failed running rescript clean")?;
-        if !exit2.success() {
-            return Ok(exit2);
-        }
```
JonoPrest, line 123: I can't remember what case we were worried about stale build 🤔, I assume it's when a file gets removed from generated (but still exists from a stale generated version)

JonoPrest (approved):
Amazing, these things can be brittle though. At least it shouldn't affect existing indexers

JonoPrest:
If we can just fix the failing cargo tests. Looks like its an expected fail now

DZakh:
Yes, I've tested several cases, including the upgrade flow from the existing version, and everything appears to be in order.

PR #607, opened 2025-06-18
Networks or ids

JasoonS:
I made this change over two months ago and they never got around to cleaning it up so here it is. Still some more cleanup to do, but it is a good start.

[Screencast from 2025-06-19 09-07-09.webm](https://github.com/user-attachments/assets/12dfc031-9d6a-4c01-a1cb-446c8a37f95e)

## Summary by CodeRabbit

- **New Features**
  - Improved network selection interface during contract import, allowing users to choose from a simplified list of supported networks.
  - Enhanced display of network names to include chain ID and a tier icon for better clarity.

- **Improvements**
  - Streamlined network selection by removing manual entry of network IDs and duplicate validation.
  - Expanded error handling and feedback for unsupported or invalid network selections.


nikbhintade:
No but planning on getting back to that after 31st Jan.

PR #610, opened 2025-06-19
Add output directory path to config

DZakh:


codegenerator/cli/src/executor/local.rs
```diff
@@ -11,21 +11,21 @@ pub async fn run_local(
     local_commands: &LocalCommandTypes,
     project_paths: &ParsedProjectPaths,
 ) -> Result<()> {
+    let config =
+        SystemConfig::parse_from_project_files(project_paths).context("Failed parsing config")?;
```
DZakh, line 15: Not cool that it requires to parse config for any local command, but I think it's worth it.

codegenerator/cli/src/config_parsing/system_config.rs
```diff
@@ -531,6 +531,44 @@ impl SystemConfig {
...
+                    // The output path is relative to the config file location
+                    let config_dir = project_paths.config.parent().ok_or_else(|| {
+                        anyhow!("Unexpected config file should have a parent directory")
+                    })?;
+
+                    let output_relative_path = PathBuf::from(output);
+                    if let Some(Component::ParentDir) =
+                        output_relative_path.components().peekable().peek()
+                    {
+                        return Err(anyhow!("Output folder must be in project directory"));
```
JonoPrest: btw there is a shorthand for this
```suggestion
                        anyhow::bail!("Output folder must be in project directory");
```

codegenerator/cli/src/executor/local.rs
JonoPrest: Hmmm, yeah I don't think its a big deal. Think these are mostly internal testing helpers anyway.

PR #630, opened 2025-07-08
Improve Envio managed db schema logic

DZakh:
This is part of the Effects Cache project. The goal is to ensure that the schema used and managed by Envio does not contain any tables unrelated to the currently running indexer. And that we have a proper initialization step, which we can use to load persisted cache from jsons.

- This also adds a safeguard error message alerting about potential data loss if the indexer started on a db schema with some existing data.
- And it makes `pnpm envio start` work without the need for `pnpm envio local db-migrate up`.
- Additionally, it reduces the likelihood of encountering a strange database state.

## Summary by CodeRabbit

* **Refactor**
  * Simplified storage initialization by removing optional parameters and streamlining logic.
  * Enforced safety checks to prevent data loss by verifying schema usage before migration.
  * Replaced hardcoded table names with dynamic references for consistency.
  * Updated network configurations by adding a new network and adjusting tier classifications.

* **Tests**
  * Revised test suites to align with updated initialization logic and removed obsolete test cases.
  * Enhanced test clarity by adjusting SQL expectations and removing deprecated parameters.

### DIFF
--- a/codegenerator/cli/npm/envio/src/Persistence.res
+++ b/codegenerator/cli/npm/envio/src/Persistence.res
@@ -15,8 +15,6 @@ type storage = {
     ~entities: array<Internal.entityConfig>=?,
     ~generalTables: array<Table.table>=?,
     ~enums: array<Internal.enumConfig<Internal.enum>>=?,
-    // If true, the storage should clear existing data
-    ~cleanRun: bool=?,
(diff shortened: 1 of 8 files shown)


moose-code (approved):
Thanks - just read through the summary and this all makes good sense as next step in persitiance. LGTM!

PR #637, opened 2025-07-14
Fix error on writing invalid utf8 string to PG

### LINKED ISSUE #446: Indexer crashing with invalid byte sequence for encoding UTF8 0x00
## Bug Description

While trying to parse an event with string parameters, the indexer crashes with:
```json 
{
 "type": "PostgresError",
 "message": "invalid byte sequence for encoding  \"UTF8\": 0x00" 
}
```

Upon further investigation it looks like parameters of type `string` are extracted from events and padded with `\x00` bytes which cannot be written to the database.

Image

The even in question is printing [this struct](https://github.com/sablier-labs/lockup/blob/e4ad4f94ea71a1a591ecb50b37b4004cdc137296/src/types/DataTypes.sol#L147) from `0xd116c275541cdbe7594a202bd6ae4dbca4578462` on Sepolia.

## Possible Fix

As temporary fix, we introduced a sanitization step before assigning the value of the event parameters to entities. By removing all the special characters (`myString.replace(/\x00/g, "")`) one can clean up the input prior to it reaching the database.

## To Reproduce

The error can be reproduced in previous versions of [`@sablier/subgraphs`](https://github.com/sablier-labs/subgraphs) by removing the sanitization step [here](https://github.com/sablier-labs/subgraphs/blob/973687bf4b2360512fd3b36f0402195b73b4b816/apps/lockup-envio/src/helpers/stream.ts#L334).

### Environment

The crash happens both locally and on the hosted environment.

**Local:**

 - Envio version: 2.12.3
 - Node version: 22.12.0
 - pnpm version: 9.12.3
 - Docker version: Client (27.4.0), Desktop (4.37.2)

**Hosted Service:**
 - Envio version: 2.12.3 (happening on earlier versions too)
 - Link to the deployed indexer: Older versions of `sablier-labs/lockup-envio` (production deployment [here](https://indexer.hyperindex.xyz/53b7e25/v1/graphql))

DZakh:
Fixes #446 

For normal cases we don't do anything, but if we encounter the error, we walkthrough all the entity string fields and try to remove the invalid sequence. This should make the case work, while not affecting performance for the events with valid data.

There was an annoying race condition, which took me much more to fix the bug than I expected 🫠

## Summary by CodeRabbit

* **New Features**
  * Added automatic removal of invalid UTF-8 null byte characters from entity history data before database insertion.
  * Introduced enhanced error handling for PostgreSQL UTF-8 encoding errors, with retry logic for affected tables.
  * Added a string utility for replacing all occurrences of a substring within a string.

* **Bug Fixes**
  * Improved error message clarity in CLI output for easier debugging.

* **Refactor**
  * Unified and streamlined database batch execution logic for entity and history operations.
  * Improved error propagation and handling during asynchronous database operations.

* **Chores**
  * Removed unused functions related to entity history row insertion.
  * Updated test code to use the new unified entity history insertion method.

### DIFF
--- a/codegenerator/cli/npm/envio/src/PgStorage.res
+++ b/codegenerator/cli/npm/envio/src/PgStorage.res
@@ -374,18 +374,27 @@ let setEntityHistoryOrThrow = (
   sql,
   ~entityHistory: EntityHistory.t<'entity>,
   ~rows: array<EntityHistory.historyRow<'entity>>,
+  ~shouldCopyCurrentEntity=?,
   ~shouldRemoveInvalidUtf8=false,
 ) => {
   rows
   ->Belt.Array.map(historyRow => {
-    let containsRollbackDiffChange =
-      historyRow.containsRollbackDiffChange->Belt.Option.getWithDefault(false)
-    let shouldCopyCurrentEntity = !containsRollbackDiffChange
     let row = historyRow->S.reverseConvertToJsonOrThrow(entityHistory.schema)
     if shouldRemoveInvalidUtf8 {
       [row]->removeInvalidUtf8InPlace
     }
-    entityHistory.insertFn(sql, row, ~shouldCopyCurrentEntity)
+    entityHistory.insertFn(
+      sql,
+      row,
+      ~shouldCopyCurrentEntity=switch shouldCopyCurrentEntity {
+      | Some(v) => v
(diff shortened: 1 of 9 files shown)


DZakh:
Ah, there's also entity history 🥲

PR #666, opened 2025-08-04
Fix greetings tests

DZakh:


JonoPrest (approved):
Cool, should we not just add a sorting filter on the query so this doesn't change?

Issue #641, opened 2025-07-16
Feature request: meta GraphQL fields for retrieving total counts


PaulRBerg:
It would be helpful to provide GraphQL fields for retrieving the total counts of entities.

Related issue in The Graph: https://github.com/graphprotocol/graph-node/issues/1309#issuecomment-3063632385

DZakh:
We don't recommend it but it's possible to expose the Hasura aggregate entity publicly by using an environment variable.

PaulRBerg:
Amazing feature

DZakh:
Also each item in array should be wrapped in double quotes. It expects a JSON format

PaulRBerg:
@DZakh I defined the `start` script like so:

```json
"start": "cd ./envio/$ENVIO_PROTOCOL && ENVIO_HASURA_PUBLIC_AGGREGATE=[\"Revenue\",\"RevenueTransaction\",\"Stream\",\"User\",\"UserTransaction\"] ts-node ./bindings/src/Index.res.js"
```

But we got build errors:

https://envio.dev/app/sablier-labs/lockup-envio/904fdd0/runtime-logs

<img width="670" height="546" alt="Image" src="https://github.com/user-attachments/assets/daef0433-8709-4697-8c91-edede9d47d42" />

PaulRBerg:
Yeah it worked with `\\\"`, thanks.

Looking forward to being able to configure the aggregate features programmatically via the YAML config! That should be much easier.

DZakh:
Nice. I'll let you know when we have a better solution 🫡

PR #676, opened 2025-08-08
Add a hack to support multiple aggregate entities on Hosted Service

DZakh:
A hacky fix for this https://github.com/enviodev/hyperindex/issues/641

We should get rid of it, when we refactor environment variables on Hosted Service.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of the `ENVIO_HASURA_PUBLIC_AGGREGATE` environment variable to better handle both array and single string formats, with clearer error messages for invalid input.


JonoPrest (approved):
Cool, wondering, do you  maybe want to make it try and parse both comma separated and and `&` so that it's backwards compatible if/when this gets fixed on the hosted service?

DZakh:
@JonoPrest , it uses union with trying to use array first. So if it's fixed on the hosted service, it will use array instead.

PR #722, opened 2025-08-28
Headless TUI

DZakh:


DZakh:
TODO:
- Use first event block as a progress start
- Fix ETA component
- Return events processed per chain when there's enough space
- Progress block number metric not always present at init start
- Show better loader when we are loading metrics

PR #770, opened 2025-09-30
Configure hasura in single bulk request - attempt to fix a race condition

DZakh:


JasoonS (approved):
Nice that you could build this :+1: Looks good to me

PR #790, opened 2025-10-22
New Templates & Changes in CLI

nikbhintade:
Added 6 new templates that users can now select directly from the CLI. Currently, all of these templates are available only in TypeScript.

CLI Updates:
- Fixed an issue where the CLI listed templates for all languages, even when only TypeScript versions existed.
- Updated logic to show templates only if that language is supported for them.
- Updated failing tests to reflect the new language-based template filtering.

## Summary by CodeRabbit

* **New Features**
  * Added multiple new init templates (topic-filtering, wildcard-indexing, on-block-handler, effects-api, factory-indexer, multichain-indexer) with sample configs, handlers, and GraphQL types.

* **Documentation**
  * Added README guides, example .env files, and .gitignore entries for each new template.

* **Chores**
  * CLI interactive init and help now assume TypeScript only; JavaScript/ReScript init options and legacy template artifacts removed.

PR #811, opened 2025-11-05
Remove JS/Res Templates

nikbhintade:
As discussed by the team, we are removing JS/Res Templates from the CLI for both EVM & Fuel.

Ref: https://github.com/enviodev/hyperindex/pull/790#pullrequestreview-3410060015

Sorry for the ambiguous branch name; the branch should've named it something better

## Release Notes

* **Breaking Changes**
  * Removed JavaScript as a supported language option. Only TypeScript and ReScript are now available for initialization.
  * Removed all JavaScript template examples, including ERC20, Greeter, and GreeterOnFuel templates.
  * Removed JavaScript event handler and test generation templates.


DZakh (approved):
Nice 🔥

PR #817, opened 2025-11-07
Allow independent Effect API fails

DZakh:
Previously if a single effect API call from a batch fails, it would cause a failure of the whole batch. With the change every effect API call can fail independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error isolation in concurrent batch operations—failures in individual calls no longer cause other concurrent calls to fail.
  * Enhanced error tracking and logging for cache loading failures to prevent silent failures.
  * Cache now stores only successful results, preventing corrupted cache entries from failed operations.

* **Tests**
  * Added end-to-end test coverage for concurrent error handling and selective caching behavior.

### DIFF
--- a/codegenerator/cli/npm/envio/src/LoadManager.res
+++ b/codegenerator/cli/npm/envio/src/LoadManager.res
@@ -7,6 +7,7 @@ module Call = {
     input: input,
     resolve: output => unit,
     reject: exn => unit,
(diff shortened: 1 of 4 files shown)


JonoPrest (approved):
Nice 🙏🏼

PR #824, opened 2025-11-12
Autoload from src/handlers

DZakh:
## Summary by CodeRabbit

* **New Features**
  * Handlers are optional and will be auto-discovered from src/handlers when not specified.
  * Per-contract handler files are generated during import; handler registration is now awaited during initialization.

* **Chores**
  * CI and package Node engine requirements tightened to Node 22.x (>=22.0.0 <=22.10.0).
  * Worktree setup now runs npm install as part of setup.
  * Persisted state no longer stores a handler file checksum.

### DIFF
--- a/codegenerator/cli/src/config_parsing/chain_helpers.rs
+++ b/codegenerator/cli/src/config_parsing/chain_helpers.rs
@@ -28,6 +28,9 @@ use crate::constants::DEFAULT_CONFIRMED_BLOCK_THRESHOLD;
 #[strum(serialize_all = "kebab-case")]
 #[repr(u64)]
 pub enum Network {
+    #[subenum(HypersyncNetwork)]
+    Ab = 36888,
+
     #[subenum(HypersyncNetwork, NetworkWithExplorer)]
     Abstract = 2741,
 
@@ -48,6 +51,9 @@ pub enum Network {
 
     #[subenum(NetworkWithExplorer)]
     ArbitrumTestnet = 421611,
+
+    #[subenum(HypersyncNetwork)]
+    ArcTestnet = 5042002,
 
     #[subenum(HypersyncNetwork, GraphNetwork, NetworkWithExplorer)]
     Aurora = 1313161554,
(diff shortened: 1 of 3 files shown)


codegenerator/cli/npm/envio/package.json
```diff
@@ -19,6 +19,9 @@
     "url": "https://github.com/enviodev/hyperindex/issues"
   },
   "homepage": "https://envio.dev",
+  "engines": {
+    "node": ">=22.0.0 <=22.10.0"
+  },
```
JonoPrest, line 24: I'm worried about this, why do we need <=22.10 again? I use v24 on my machine for eg

DZakh: Mocha starts attempting to run tests in esm, which breaks in some cases

DZakh: The reasoning that this PR will go to v3.0.0-alpha.0, so it's fine to have the limitation, since one of the next alphas will bring ESM only mode, and lift the restriction of node.js <=22.10

DZakh: There are no problmes with running the indexer, only when running mocha

JonoPrest: Ah cool  👍🏼 so this will not be the latest version for some time

DZakh: Yes, until ESM support

codegenerator/cli/templates/dynamic/codegen/src/TestHelpers_MockDb.res.hbs
```diff
@@ -329,7 +329,7 @@ and makeProcessEvents = (mockDb: t, ~chainId=?) => async (
     ) {
       registrations
     } else {
-      Generated.registerAllHandlers()
+      await Generated.registerAllHandlers()
```
JonoPrest, line 332: I seem to remember if we made this async it causes problems for the testing framework. Does that still work fine?

codegenerator/cli/templates/static/greeter_template/typescript/test/test.ts
```diff
@@ -1,5 +1,7 @@
 import assert from "assert";
-import { TestHelpers, User } from "generated";
+import { TestHelpers } from "generated";
+import type { User } from "generated";
```
JonoPrest, line 3: Is this because we're using interfaces now or something?

JonoPrest (approved):
Cool Dmitry, 

Looking good :) exciting to be thinking about v3

codegenerator/cli/templates/dynamic/codegen/src/TestHelpers_MockDb.res.hbs
DZakh: Yes, I changed the testing framework some time ago to support this case

codegenerator/cli/templates/static/greeter_template/typescript/test/test.ts
DZakh: No, this was an issue when I ran mocha with node >=22.10 and it used esm module resolution and didn't know what a type is.

PR #869, opened 2025-12-18
Use `0x${string}` for Address type + add S.address

DZakh:
## Summary by CodeRabbit

* **New Features**
  * Address type is now part of the public API and an address schema is exposed.

* **Type System Improvements**
  * Address values are enforced as 0x-prefixed hex strings, tightening address validation across the SDK.

* **Tests**
  * Tests and event typings updated to reflect the stricter address format and schema changes.


codegenerator/cli/npm/envio/src/Types.ts
```diff
@@ -1,6 +1,6 @@
 export type Invalid = never;
 
-export type Address = string;
+export type Address = `0x${string}`;
```
JonoPrest, line 3: Will this have any impact if say a function expects an address as a string but now it has a template literal type?

DZakh: No impact if it expects a string, since it's a subset of a string. But if somebody want to assign a general string to Address type, it'll complain.

PR #871, opened 2025-12-18
Add `envio config list`

DZakh:
An idea to use later to get config as a JSON on a hosted service, as well as for generating. I want to consume json instead of generating code in rust.

I also want to add entities and events types using a superset of JSON schema.


DZakh:
I'm actually not 100% sure we need the API. Maybe generating config.json will be enough

JonoPrest (approved):
Cool 👍🏼

PR #914, opened 2026-01-16
Generate Entity types without dependency on ReScript

DZakh:
## Summary by CodeRabbit

* **New Features**
  * Added Status enum and RelatedEntity type to test schema.
  * Added derived-field tracking for entity parameters to improve generated types.

* **Updates**
  * Generated entity access now uses bracket-notation (Entities["X"]) across handlers, exports, and templates.
  * Array and enum TypeScript output now uses readonly wrappers and bracket-style enum references (Enums["Name"]).
  * Updated imports to expose generated and public entity types separately; adjusted per-entity method signatures accordingly.


codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -154,17 +154,20 @@ impl GraphQlEnumTypeTemplate {
...
-    pub res_type: TypeIdent,
+    #[serde(rename = "res_type")]
+    pub field_type: TypeIdent,
     pub is_entity_field: bool,
     pub is_indexed_field: bool,
     ///Used to determine if you can run a where
     ///query on this field.
     pub is_queryable_field: bool,
+    /// Whether this field is derived from another entity (not stored in DB).
+    pub is_derived_field: bool,
```
JonoPrest, line 165: Very small nitpick. 

If we don't need this as a flag for handlebars then I think its more idiomatic to make this a method on the struct that calls the underlying is_derived_from() fn on field_type.

Or just call it directly on the field_type where its used.

DZakh: Accepted. But it's quite annoying to have different types for handlebars and rust codegen for now. We should definetely have a good refactoring after handlebars is removed :)

PR #922, opened 2026-01-20
Defer config validation until indexer runs

DZakh:
## Release Notes

* **New Features**
  * Enhanced source configuration system supporting multiple blockchain ecosystems (EVM, Fuel, SVM) with ecosystem-specific settings
  * Event signatures now explicitly managed in contract configurations
  * Structured RPC configuration with improved metadata handling

* **Refactor**
  * Reorganized configuration architecture for improved source setup and lazy initialization

### DIFF
--- a/codegenerator/cli/npm/envio/index.d.ts
+++ b/codegenerator/cli/npm/envio/index.d.ts
@@ -349,6 +349,7 @@ type IndexerConfigTypes = {
     contracts?: Record<string, {}>;
   };
   svm?: { chains: Record<string, { id: number }> };
+  entities?: Record<string, object>;
 };
 
 // Helper: Check if ecosystem is configured in a given config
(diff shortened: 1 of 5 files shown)


codegenerator/cli/templates/dynamic/codegen/src/Generated.res.hbs
```diff
@@ -22,11 +23,9 @@ let makeGeneratedConfig = () => {
         },
         {{/each}}
       ]
-      let chain = ChainMap.Chain.makeUnsafe(~chainId={{chain_config.network_config.id}})
       {
         Config.id: {{chain_config.network_config.id}},
         contracts,
-        sources: {{chain_config.sources_code}}
```
DZakh, line 29: Previously we initialized HyperSyncSoure.make here which would validate an ENVIO_API_TOKEN in env variables. This broke hosted service, since the config is also read during db initialization which doesn't have the same env variables we have during indexer run. I decided to fix it by removing more runtime code from generated and passing simple config instead. Now the sources are initialized only when indexer is running, not on the config import

DZakh: See https://github.com/enviodev/hyperindex/pull/922/commits/3008a8fe250049abc9252dcb0963e0eb4f1c6a8b

JonoPrest: Ok cool

codegenerator/cli/src/hbs_templating/codegen_templates.rs
```diff
@@ -96,6 +93,32 @@ struct InternalSvmConfig {
...
+    backoff_millis: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fallback_stall_timeout: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    query_timeout_millis: Option<u32>,
+}
+
+fn is_empty_vec<T>(v: &Vec<T>) -> bool {
+    v.is_empty()
+}
```
JonoPrest: seems a bit redundant

PR #937, opened 2026-02-06
Prefix and scope the internal 'core' envio platform specific packages

JasoonS:
Changes looked good (AI helped me with a command to review it):
```
# 1. Test template substitution
cd codegenerator/cli/npm
export node_pkg="core-linux-x64" node_os="linux" node_arch="x64" version="2.0.0-test"
envsubst < package.json.tmpl > /tmp/test.json && cat /tmp/test.json | grep '"name"'
# Should show: "name": "@envio-dev/core-linux-x64"

# 2. Verify binary exists
ls -la ../../target/release/envio 2>/dev/null || echo "Build first with: cd ../.. && cargo build --release"

# 3. Check main package template
cd envio
envsubst < package.json.tmpl > /tmp/main.json && cat /tmp/main.json | grep -A5 optionalDependencies
# Should show @envio-dev/core-* packages
```

## Release Notes

* **Chores**
  * Updated NPM package naming scheme to use scoped packages under @envio-dev namespace instead of non-scoped packages.
  * Modified release workflow and binary resolution paths to align with the new package naming convention.
  * Updated optional dependencies to reference the new scoped package names across the build infrastructure.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -251,7 +251,7 @@ jobs:
           if [ "${OS}" = "windows-2022" ]; then
             bin="${bin}.exe"
           fi
-          cp "../../target/${TARGET}/release/${internal_name}" "${node_pkg}/bin"
+          cp "../../target/${TARGET}/release/${bin}" "${node_pkg}/bin"
           cp ../README.md "${node_pkg}"
           cd "${node_pkg}"
           npm publish --access public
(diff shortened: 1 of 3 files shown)


DZakh:
Sorry for the hustle, but I'd like to postpone it a little bit. I want to start using napi for the envio-cli package, so I'm not sure how it'll require changing the package name. I don't want to pollute npm with too many packages like this

PR #976, opened 2026-02-20
Add validation for type declarations and schema generation

moose-code:
## Summary
This PR adds comprehensive validation to type declarations and schema generation, converting several TODO comments into implemented validation logic. The changes improve error handling by detecting invalid configurations early and providing clear error messages.

## Key Changes

- **TypeDeclMulti validation**: 
  - Changed `new()` to return `Result<Self>` instead of `Self`
  - Added validation to reject empty type declaration lists
  - Added duplicate name detection among type declarations

- **TypeDecl validation**:
  - Changed `new()` to return `Result<Self>` instead of `Self`
  - Added validation to detect duplicate type parameters

- **Topological sort improvements**:
  - Changed `to_rescript_schema()` to return `Result<String>` instead of `String`
  - Added cycle detection in dependency resolution to prevent infinite loops
  - Distinguished between external dependencies (allowed) and internal cyclic dependencies (error)
  - Improved algorithm to detect when no progress is made during sorting

- **Schema validation**:
  - Implemented `check_field_names_for_proto()` to detect `__proto__` field names that could cause prototype pollution issues
  - Integrated the check into the schema validation pipeline

- **Contract event validation**:
  - Implemented duplicate event name detection in `Contract::new()`
  - Provides clear error messages when duplicate event names are found

- **Updated all call sites**: Modified all test cases and production code to handle the new `Result` return types with `.unwrap()` or `?` operators as appropriate

- The topological sort now tracks both declared type names and dependencies, allowing it to distinguish between missing internal types (error) and external type references (allowed)
- Cycle detection uses a progress-tracking approach: if no new types are added in an iteration, a cycle must exist
- All validation errors use `anyhow!()` for consistent error handling across the codebase

https://claude.ai/code/session_01BwrGigSvKsk4MR43QhkSrx

## Summary by CodeRabbit

* **New Features**
  * Added schema validation to prevent configuration issues with restricted field naming patterns.

* **Bug Fixes**
  * Improved duplicate event name detection with detailed error reporting.
  * Enhanced error propagation in type definition handling.
  * Strengthened type schema validation with cycle-dependency detection.


DZakh (changes requested):
This looks good, but I'd like to have a test for every case which was fixed. I see there are tests in type_schema.rs, but it uses internal API which is not very convenient to review + more subject to break on any internal change. Ideally to have the tests accepting schema.graphql code as a string, parsing it and verifing that it either works correctly, or we get a specific error message 👍

PR #977, opened 2026-02-20
Optimize buffer merging with O(n+m) merge instead of sort

moose-code:
## Summary
This PR optimizes the buffer update logic in FetchState by implementing an efficient merge algorithm for combining sorted arrays, reducing time complexity from O((n+m) log(n+m)) to O(n+m) in the common case.

## Key Changes
- **Added `mergeSortedBuffers` function**: Implements a standard two-pointer merge algorithm for combining two sorted arrays in linear time. Includes fast paths for cases where one array entirely precedes the other.
- **Updated `updateInternal` signature**: Added optional `newSortedItems` parameter as an alternative to `mutItems`, allowing callers to provide pre-sorted items for efficient merging.
- **Optimized `handleQueryResult`**: Changed to sort incoming items once and merge with the existing buffer using `mergeSortedBuffers`, rather than concatenating and sorting the entire combined array.
- **Enhanced buffer update logic**: The buffer assignment now handles four cases:
  - New sorted items only: merge with existing buffer
  - Both mutItems and new sorted items: sort mutItems first, then merge
  - MutItems only: sort as before (fallback for unsorted data)
  - Neither: use existing buffer unchanged

## Implementation Details
- The merge algorithm uses unsafe array access (`Js.Array2.unsafe_get/set`) for performance-critical operations
- Fast path checks compare the last element of one array with the first element of the other to detect non-overlapping ranges
- The solution maintains backward compatibility by keeping `mutItems` as an option for cases where data sources return unsorted items
- Time complexity improvement: from O((n+k) log(n+k)) to O(k log k + n) where k is the number of new items and n is the buffer size

https://claude.ai/code/session_01HprRgtzm8JjstbRCcL1xT3

### DIFF
--- a/packages/envio/src/FetchState.res
+++ b/packages/envio/src/FetchState.res
@@ -539,6 +539,66 @@ let compareBufferItem = (a: Internal.item, b: Internal.item) => {
   }
 }
 
+/**
+Merges two sorted arrays into a single sorted array in O(n+m) time.
+Both input arrays must be sorted by (blockNumber, logIndex) ascending.
+Includes fast paths for the common case where one array entirely precedes the other.
+*/
+let mergeSortedBuffers = (sortedA: array<Internal.item>, sortedB: array<Internal.item>) => {
+  let lenA = sortedA->Array.length
+  let lenB = sortedB->Array.length
+  if lenA === 0 {
+    sortedB
+  } else if lenB === 0 {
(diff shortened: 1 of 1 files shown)


DZakh (changes requested):
Looks scary. I'd like to have a benchmark for changes like this first 😅

PR #1344, opened 2026-06-24
Replace SVM field selection boolean with explicit field lists

DZakh:
## Summary

Refactors SVM instruction field selection from a boolean-or-list union type to explicit per-field selection via enums and lists. This improves type safety, enables per-instruction compile-time errors for unselected fields, and simplifies the configuration API.

## Key Changes

- **Config schema**: Replaced `FieldSelectionValue` (bool | list) with:
  - `SvmTransactionField` enum for transaction fields (10 variants: transactionIndex, signatures, feePayer, etc.)
  - `transaction_fields: Option<Vec<SvmTransactionField>>` (list of selected fields)
  - `log_fields: Option<bool>` (simplified to boolean)
  - `token_balance_fields: Option<bool>` (simplified to boolean)

- **Field resolution**: Removed validation against `SVM_TRANSACTION_FIELDS` constant; now uses enum's `Display` trait for camelCase serialization. Deduplicates selected fields in declared order.

- **TypeScript codegen**: Added `svm_transaction_ts_type()` function that generates per-instruction transaction types where:
  - Selected fields get their real TS type
  - Unselected fields become `FieldNotSelected<...>` to produce compile errors pointing users to the `field_selection` config knob
  - Matches the EVM block/transaction UX

- **Type parameters**: Extended `SvmInstruction<Params, Tx>` with a `Tx` type parameter (defaults to `SvmTransaction`) so handlers receive the per-instruction transaction shape from codegen.

- **Config examples**: Updated all test configs and scenario files to use explicit field lists (e.g., `transaction_fields: [signatures, feePayer]`) instead of `transaction_fields: true`.

## Implementation Details

- Removed per-field validation error for `transaction_fields` in `validation.rs` (no longer needed; enum enforces valid fields at parse time)
- Simplified `resolve_svm_transaction_fields()` to return `Vec<String>` without `Result` wrapper
- Updated JSON schema to reflect new field types
- Preserved field order and deduplication logic during resolution

https://claude.ai/code/session_015C9ndPs24AHA7CTFK6CpuE


packages/envio/index.d.ts
```diff
@@ -997,9 +997,11 @@ export type SvmTokenBalance = {
   readonly postAmount?: string;
 };
 
-/** Parent transaction surfaced when an instruction's
- * `include_transaction` flag is `true`. */
+/** Full parent-transaction shape. Per-instruction handler types narrow this to
+ * the fields selected via `field_selection.transaction_fields`; this standalone
+ * type is the fallback when the program table hasn't been generated. */
 export type SvmTransaction = {
```
DZakh: I think we shouldn't have a fallback like this and always use a generated type. Check how EVM ecosystem does this.

scenarios/svm_flow_xray/src/handlers/flow.ts
```diff
@@ -43,13 +44,29 @@ type NodeArgs = {
...
 type MapArgs = (params: SvmInstructionParams) => NodeArgs;
-type FlowHandler = (a: { instruction: SvmInstruction; context: FlowContext }) => Promise<void>;
+
+// Every handler is registered through the shared factories below, which take the
+// generic `SvmInstruction` — so the per-instruction `FieldNotSelected` compile
+// guard (which fires on `indexer.onInstruction`'s inferred argument) doesn't
+// reach these bodies. `FlowInstruction` pins the exact transaction fields the
+// handlers read, so dropping one from config.yaml's `transaction_fields` is a
+// type error here instead of a silent `undefined` at runtime.
+type FlowTransaction = {
```
DZakh: Don't create your own types and scenarios and reuse types exported from envio package.

packages/cli/src/config_parsing/chain_helpers.rs
```diff
@@ -59,6 +59,7 @@ pub enum Network {
     #[subenum(GraphNetwork, NetworkWithExplorer)]
     AuroraTestnet = 1313161555,
 
+    #[subenum(HypersyncChain)]
```
DZakh: Remove

scenarios/svm_flow_xray/src/handlers/flow.ts
```diff
@@ -1,223 +1,397 @@
...
+    program: "Kamino",
+    ixName: instruction.params?.name ?? "depositReserveLiquidityAndObligationCollateral",
+    programId: instruction.programId,
+    isInner: instruction.isInner,
+    slot: instruction.block.slot,
+    addr: instruction.instructionAddress,
+    txSig: tx?.signatures[0],
+    feePayer: tx?.feePayer,
+    success: tx?.success,
+    fee: tx?.fee,
```
DZakh: Why is this optional? Confirm that it compiles correctly with TypeScript after codegen.

DZakh: Confirmed — it compiles. I bootstrapped the scenario's `envio codegen` + `tsc -p tsconfig.json` this session and it passes against the generated types.

`fee: tx?.fee` is optional for two independent reasons:
1. `tx?.` — `instruction.transaction` is typed `transaction?: Tx` (`index.d.ts`), since the transaction object is absent when an instruction selects no transaction fields. This scenario selects fields, so it's present at runtime, but the `?.` is required by the type.
2. `tx?.fee` resolves to `bigint | undefined` because `fee` is generated as a **nullable** selected field. In the generated transaction shape only `signatures` is non-nullable (`readonly string[]`); `feePayer`, `success`, `fee`, and `computeUnitsConsumed` are all `T | undefined`. That flows into the nullable `fee` entity column.

If the intent is that `fee`/`feePayer`/`success` are always present once selected (only `computeUnitsConsumed` is genuinely absent on older txs), that's a one-line change to their optionality in `svm_transaction_field_specs` — happy to make it if you want them non-nullable.

---

PR #1354, opened 2026-06-26
fix: support earlier registrationBlock for dynamic contracts (addresses #1188)

### LINKED ISSUE #1188: Skipping contract registration: Contract address is already registered at a later block number

duckling69:
Hey folks,

Been running into the "Skipping contract registration: Contract address is already registered at a later block number" warning in a few setups with dynamic contracts. The root seems to be that we key on address in indexingAddresses and just take whatever registration we see first. Because partitions are concurrent and the items coming back from a source query aren't strictly guaranteed to be in block order (plus within a batch), a later sighting could win and lock in a too-high effectiveStartBlock for that address. Then we skip the earlier one and the persisted row in envio_addresses ends up with the wrong (later) block.

This patch makes us prefer the *earliest* registrationBlock we encounter for an address. When we see a strictly earlier sighting we update the record, log a correction at info level instead of the scary warning, and keep the DC attached so it gets persisted with the right (early) block. For stuff we already had indexed we also lower the latestFetchedBlock on the owning partition(s) so we actually fetch the missing earlier range (the client-side filters already drop pre-effective events for other addresses in the same partition). Added a bit of bookkeeping with chosenDcItemByAddress so within a single register call the min wins even if the array isn't perfectly sorted.

Only touched the three files needed for the actual fix + the test updates that were exercising the old (broken) expectations.

Built clean with rescript (124 modules) and the FetchState registerDynamicContracts tests now pass with the new behavior.

Fixes #1188. Would really appreciate a review, especially on the backfill approach (lowering LFB vs. emitting an explicit short gap partition) and anything reorg-related I might have missed. Happy to tweak.

Thanks!

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dynamic-contract tracking when the same address is rediscovered with an earlier start block, ensuring the earliest start is used consistently.
  * Corrected the persisted registration block so saved results reflect the earliest valid registration.
  * Improved resilience to out-of-order batch/partition results by adjusting subsequent fetch behavior to recheck corrected earlier ranges.
* **Tests**
  * Expanded `registerDynamicContracts` test coverage for earlier-registration corrections, within-batch ordering, and successive out-of-order updates.

### DIFF
--- a/packages/envio/src/FetchState.res
+++ b/packages/envio/src/FetchState.res
@@ -1009,33 +1009,25 @@ let registerDynamicContracts = (
   // Might contain duplicates which we should filter out
   items: array<Internal.item>,
 ) => {
-  // === Batch dynamic-contract registration invariant (senior dev note) ===
-  // For every address we process in this call we must guarantee:
-  //   - At most one DC remains attached to any item.
-  //   - The attached DC (if any) comes from the sighting with the *smallest*
-  //     registrationBlock seen for that address *anywhere in this call*.
-  // This single value is what ends up in indexingAddresses (for effectiveStartBlock
-  // filtering) and in the envio_addresses row (for persistence and restart).
+  // The logic below ensures that for any address processed in one call to this
+  // function, at most one dynamic contract registration survives (the one whose
+  // registrationBlock is the smallest seen anywhere in the items passed to this
+  // call). That value is used both for the effectiveStartBlock that controls
+  // filtering and for the row that gets written to envio_addresses.
   //
-  // Because a single registerDynamicContracts call can receive items from multiple
-  // partitions (and the array order is not guaranteed), we maintain two pieces of
-  // per-address "best so far" state for the current call:
-  //   registeringAddresses / correctedAddresses  – the best indexingAddress seen
-  //   chosenDcItemByAddress                       – which item currently "owns" that best DC
+  // A single call can contain items coming from several partitions, and the order
+  // of items inside the array is not guaranteed to be sorted by block number.
+  // We therefore keep per-address "best so far" state for the duration of the call
+  // and, when a strictly earlier sighting is found later in the array, we remove
+  // the DC that was attached to an earlier item. This makes the final result
+  // independent of the order in which the items happened to arrive.
   //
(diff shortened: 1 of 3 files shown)


duckling69:
Hey, just opened this for #1188.

The warning was firing because we were treating the first registration sighting we saw for an address as authoritative. With partitions running in parallel and responses not being strictly ordered by block, a later registration could get processed first, and then any earlier one would get skipped with that exact message. effectiveStartBlock ended up too high and the row we persisted to envio_addresses was also the wrong one.

What I did:
- In registerDynamicContracts we now take the min when we see an earlier sighting for the same address (both against what we already have in indexingAddresses and within the current batch of items).
- On a correction we update the record, log at info instead of warning, and make sure the DC stays on the item so the early block makes it to setBatchDcs.
- Added a little chosenDcItemByAddress thing so that even if the items array in one call isn't perfectly sorted, the earliest sighting wins and later ones in the same batch get their DC spliced off.
- For addresses we already knew about, we lower latestFetchedBlock on the partitions that own them so we actually go back and fetch the missing range.
- In setBatchDcs we now prefer the registrationBlock that was on the dc (the one from the actual sighting) rather than blindly taking the processing event's block.

Only the bits that actually change the behavior + the tests that were asserting the old behavior. Build is clean and the specific tests for this path pass now.

Would love eyes on it, especially around whether lowering the partition LFB is the right way to drive the backfill or if we should be creating a dedicated short gap partition with a mergeBlock. Also any reorg/rollback gotchas I might have missed.

Happy to adjust based on feedback. Thanks!

DZakh:
Hi, thanks for contributing. Unfortunately, I think lowering LFB won't work in this case, since partition might already be fetching when we get a new registration + it might have other addresses which already got logs for the block range, so reducing LFB might result in duplicated logs. Probably the right approach here will be to create a new partition with a start and end block to catch up to the already existing partition.

But I'm letting you know that I'm currently working on version 3.3.0, which changes the way we store addresses a lot, so the PR will have many conflicts there. I suggest you to stop working on it in the meantime. I understood that this is a problem for you, so I'll try to include support for the case in the internal redesign I work on for 3.3.0 🙏

PR #1411, opened 2026-07-13
Validate block-store responses before retry

DZakh:
## Summary

- validate page-internal and cross-response block hash consistency in the Rust block store
- retry inconsistent block responses through the existing `onReorg` path while preserving request statistics
- move EVM and SVM rollback hash pagination into Rust, including parent-slot and parent-blockhash validation for skipped SVM slots
- materialize block data from persistent block stores across source implementations

## Why

A response can contain block hashes from inconsistent backend instances or forks. Keeping response conflict metadata and cross-store comparison inside `BlockStore` distinguishes an invalid response from a confirmed chain reorg and keeps block-only data out of ReScript.

## Validation

- `cargo check -p envio --lib`
- `cargo fmt --all -- --check`
- `pnpm build` in `packages/envio`
- `pnpm build` in `scenarios/test_codegen`
- focused Vitest suite: 67 tests passed
- Rust unit suites: BlockStore 21, EVM HyperSync 8, SVM query 4, request stats 1
- `git diff --check`


packages/cli/src/evm_hypersync_source/mod.rs
```diff
@@ -86,20 +106,77 @@ impl EvmHypersyncClient {
...
+        if from_block < 0 {
+            return Err(map_err(anyhow::anyhow!(
+                "block numbers must be non-negative"
+            )));
+        }
+        if to_block - from_block > 1_000 {
+            return Err(map_err(anyhow::anyhow!(
+                "Invalid block data request. Range of block numbers is too large. Max range is 1000. Requested range: {from_block}-{to_block}"
+            )));
+        }
```
DZakh: I don't see reason for that imitation.

packages/cli/src/evm_hypersync_source/mod.rs
```diff
@@ -86,20 +106,77 @@ impl EvmHypersyncClient {
...
+                field_selection: query::FieldSelection {
+                    block: Some(vec![BlockField::Number, BlockField::Hash]),
+                    ..Default::default()
+                },
+                ..Default::default()
+            };
+            let started = Instant::now();
+            let response = self.get_raw(query).await;
+            request_stats.push(RequestStat {
+                method: "getBlockHashes".to_string(),
```
DZakh: queryBlockHashes

packages/cli/src/svm_hypersync_source/mod.rs
```diff
@@ -176,6 +185,102 @@ impl SvmHypersyncClient {
...
+            if next_slot > to_slot || next_slot <= cursor {
+                break;
+            }
+            cursor = next_slot;
+        }
+
+        // A successor can only prove a missing suffix. Interior omissions
+        // already have a returned descendant, so another later block cannot
+        // repair their parent link.
+        if aggregate.missing_hashes(block_numbers).contains(&to_slot) {
```
DZakh: Explain why this is needed

packages/cli/src/svm_hypersync_source/mod.rs
```diff
@@ -176,6 +185,109 @@ impl SvmHypersyncClient {
...
+            aggregate.missing_hashes(block_numbers).contains(&to_slot);
+
+        // An interior skipped slot has a later block inside the bounded
+        // response whose (parent_slot, parent_blockhash) link proves the gap.
+        // A skipped `to_slot` has no such descendant in [from_slot, to_slot],
+        // so fetch exactly one successor. Its parent link either proves that
+        // the upper slot was skipped or leaves it missing as an invalid
+        // response; without this lookup a valid skipped upper slot would be
+        // retried forever.
+        if upper_slot_needs_successor_proof {
```
DZakh: If HyperSync didn't return a slot in the range between from slot and response slot then it means it is trully missing.

PR #1417, opened 2026-07-14
Add Rust-native Hasura-compatible GraphQL server

DZakh:
## Summary

Adds `envio serve`, a Rust-native GraphQL server for indexed Postgres data with a Hasura-compatible read API.

The branch covers the complete serve stack: CLI/configuration, Postgres catalog introspection, GraphQL schema generation, validation and SQL compilation, HTTP execution, WebSocket subscriptions, TLS/auth behavior, production hardening, and a differential test harness backed by real Hasura responses.

The WebSocket follow-up hardens subscriptions against denial-of-service workloads and reduces duplicate database polling.

## Why

Serving indexed data currently depends on running a separate Hasura deployment. A native server removes that operational dependency while preserving the API clients already use.

The WebSocket follow-up is necessary because the original subscription implementation had unbounded operations per socket, one Postgres poll per subscriber, synchronized polling, no initialization deadline, and unbounded outbound buffering. A single client could create enough subscriptions to saturate the shared database pool or consume process memory.

### Hasura-compatible GraphQL server

- Adds the `envio serve` CLI command and environment-based configuration.
- Introspects generated Postgres schemas and builds Hasura-shaped GraphQL types.
- Implements queries, filters, relationships, ordering, pagination, distinct-on, aggregates, introspection, variables, aliases, fragments, and Hasura-compatible validation errors.
- Executes independent root fields concurrently.
- Supports admin-secret and role-based authorization behavior.
- Supports Postgres TLS modes, startup retry, structured logging, graceful shutdown, and explicit HTTP load protection.
- Retains compatibility with legacy Envio v2.21.5 project configuration.

### HTTP and WebSocket API

- Serves the existing `/v1/graphql` API.
- Supports both WebSocket subprotocols used by Hasura:
  - modern `graphql-transport-ws`
  - legacy `graphql-ws`
- Implements ordinary live queries and cursor-based streaming subscriptions.
- Preserves initial result delivery, changed-result suppression, modern/legacy message envelopes, authentication during upgrade or `connection_init`, ping/pong, transient database retry, and graceful shutdown.

### WebSocket performance and resilience

- Caps concurrent WebSocket connections, operations per connection, and operations globally.
- Groups identical live queries by compiled SQL, bound parameters, and role, polls once, and fans the result out to every subscriber.
- Compiles ordinary live-query SQL once instead of once per poll.
- Adds a configurable polling interval and stable jitter.
- Caps concurrent WebSocket database polls below the Postgres pool size so HTTP traffic retains database capacity.
- Adds a non-extendable `connection_init` timeout.
- Replaces the unbounded outbound channel with a bounded queue.
- Makes WebSocket frame/message and HTTP body limits explicit.
- Adds Envio configuration variables plus fallbacks for relevant Hasura environment variables.

Exact-query cohorts deliberately do not combine different variable sets. This captures identical-query fan-out with a simple correctness boundary; it is not Hasura's deeper cross-variable SQL multiplexing. Streaming subscriptions remain independent because each subscriber owns a mutable cursor.

## User-facing impact

Existing GraphQL documents, variables, response shapes, authentication headers, the `/v1/graphql` endpoint, and the default one-second live-query cadence remain compatible.

Clients only observe new behavior at safety boundaries:

- HTTP 503 when the global WebSocket connection limit is full.
- close 1008 when one socket exceeds its operation limit.
- close 1013 when the global operation limit is full.
- close 4408 when `connection_init` misses its deadline.
- close 4429 for duplicate initialization on the modern protocol.
- HTTP 413 or WebSocket closure for oversized payloads.
- disconnect when a client stops reading and fills its bounded response queue.

Temporary 503/1013 responses should be retried with backoff. A 1008 response requires the client to use fewer operations per socket.

## Protocol compatibility

The names are easy to confuse:

- `graphql-transport-ws` is the current protocol. It uses `subscribe`/`next`, protocol-level `ping`/`pong`, and standardized WebSocket close codes. It is implemented by the modern npm package named `graphql-ws`.
- `graphql-ws` is the legacy wire subprotocol associated with the deprecated `subscriptions-transport-ws` ecosystem. It uses `start`/`data`/`stop` and legacy keepalive/error messages.

Both remain supported so existing Hasura clients can migrate independently.

## Validation

- Rust formatting passes: `cargo fmt --all -- --check`.
- Targeted Docker-backed tests pass for:
  - exact live-query cohort deduplication
  - global connection limits
  - per-connection and global operation limits
  - initialization timeout behavior under ping traffic
  - explicit WebSocket message limits
- Rust 1.97 Clippy passes with `-D warnings`.
- The full CI test command passes locally with Docker and network access: 452 passed, 0 failed, 2 ignored.
- The live HyperSync health suite passes after checking every supported endpoint.
- The branch includes a 621-case Hasura differential read-surface corpus, WebSocket protocol compatibility tests, robustness tests, benchmarks, and concurrent soak/load coverage.

## Review notes

This is a large diff because the differential corpus records Hasura oracle snapshots. Most files are test fixtures; the implementation is concentrated under `packages/cli/src/serve`.


.github/workflows/build_and_verify.yml
```diff
@@ -9,6 +9,10 @@ on:
   push:
     branches:
       - main
+  # Nightly: re-record the differential snapshots against live Hasura and
+  # fail on drift (differential-snapshot-drift job below).
+  schedule:
+    - cron: "0 3 * * *"
```
DZakh: Let's not do this

DZakh: I want all the runs be triggered by pull requests

PR #1421, opened 2026-07-14
Centralize config parsing tests around YAML

DZakh:
## What changed

- Added an in-memory config source that accepts YAML, schema text, environment variables, and virtual ABI/IDL files while reusing the production Rust parser.
- Exposed the string parser through NAPI/ReScript and added `MockIndexerConfig.parseYaml` for high-level tests.
- Consolidated user-facing config parsing and validation cases into `ConfigYaml_test.res`, removing redundant focused Rust and `Config.res` parsing tests.
- Updated column-name tests to build their config through YAML.
- Made RPC field-selection validation chain- and contract-aware for mixed RPC/HyperSync projects.
- Improved several user-facing validation errors and validates SVM public keys by decoding exactly 32 bytes.

## Why

Config behavior was tested through several lower-level entry points that did not match how users provide `config.yaml`. The parser also depended directly on project filesystem state, making high-level isolated tests difficult.

The mixed-source field-selection check additionally used one project-wide RPC flag, so an unrelated RPC chain could incorrectly reject fields used only by a HyperSync contract.

## Impact

Config tests now exercise the full YAML → Rust system config → public config → ReScript config path. Tests can provide schema, environment, and referenced files explicitly and run independently of the current directory.

The normal project parser remains filesystem-backed and shares the same internal implementation and NAPI error serialization path.

## Validation

- `cargo clippy -p envio --lib`
- `cargo test -p envio config_parsing --lib` — 122 passed, 1 ignored
- `pnpm exec rescript`
- Focused Config YAML/Config suites — 120 passed
- Column naming integration suite — 14 passed
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added in-memory YAML configuration parsing through the CLI and ReScript APIs.
  * Supports supplying schemas, environment variables, virtual files, and parsing options without local project files.
  * Added support for resolving virtual ABI and IDL files during configuration parsing.

* **Bug Fixes**
  * Improved Solana public-key validation to verify decoded key length.
  * Improved configuration validation errors with clearer context for invalid addresses and block ranges.

* **Tests**
  * Expanded end-to-end coverage for configuration parsing, validation, interpolation, schemas, and virtual files.


packages/cli/Cargo.toml
```diff
@@ -58,6 +58,7 @@ convert_case = "0.6.0"
 bollard = "0.20"
 futures-util = "0.3"
 tar = "0.4"
+bs58 = "0.5.1"
```
DZakh: What is this and why is it needed?

DZakh: `bs58` is used by `is_valid_solana_pubkey` to decode user-supplied program IDs and account-filter values, so we reject strings that are valid base58 syntax but do not decode to a 32-byte Solana key. It is already transitively present through `hypersync-client-solana`, so this direct declaration does not add another crate to the build; Rust requires it to be declared directly because we call it from this crate. I moved it beside the Solana dependencies, documented the reason, and changed the validator to decode into a fixed 32-byte buffer without allocating.

PR #1483, opened 2026-07-24
Add client-side address filtering for high-volume contracts

DZakh:
## Summary

Implements automatic switching of high-volume contracts to client-side (wildcard) address filtering. When a dynamic contract's registered address count exceeds a configurable threshold, the indexer switches it to fetch logs address-free from the RPC/HyperSync source, then applies address filtering in JavaScript. This reduces server-side partition fragmentation and query overhead for contracts with many registered addresses.

## Key Changes

- **Wildcard contract tracking**: Added `wildcardContracts` set to `OptimizedPartitions.t` to track contracts switched to client-side filtering. These contracts are queried address-free and their registrations are collapsed into a single address-free partition.

- **Selection metadata**: Extended `selection` type with optional `clientSideFilteredContracts` array to communicate which contracts use client-side filtering to the source layer.

- **Partition collapsing**: New `collapseWildcardContracts` function merges address-free partitions and single-contract dynamic partitions (for wildcard contracts) into one address-free partition at their minimum frontier. Re-fetching the overlap is deduped by `mergeIntoBuffer`.

- **Dynamic contract switching**: In `registerDynamicContracts`, contracts crossing the `wildcardAddressThreshold` are added to `wildcardContracts` and their new addresses trigger a backfill query instead of a standing partition.

- **IndexingAddresses restructuring**: Changed from flat `dict<indexingAddress>` to nested `dict<dict<indexingAddress>>` (grouped by contract name) to support O(1) per-contract address counting and provide the contract-scoped dict needed by `clientAddressFilter`.

- **Source integration**: Updated Rust selection builders and decoders (HyperSync, RPC) to:
  - Accept `ClientFilteredContracts` parameter
  - Build address-free selections for client-filtered contracts
  - Treat `ContractAddresses` topic markers as match-any when `force_wildcard` is set
  - Route client-filtered registrations as wildcards (accept any emitter)

- **Configuration**: Added `wildcardAddressThreshold: option<int>` to `FetchState.t` (gated to EVM chains only) and `WILDCARD_ADDRESS_THRESHOLD` environment variable.

- **Query response handling**: Refactored `handleQueryResponse` to handle partitions absorbed into wildcard partition (returns early if partition no longer exists).

- Address strings are globally unique across contracts; conflicting registrations are rejected before reaching `IndexingAddresses`.
- The `clientAddressFilter` (generated JS) gates each item by registered address + `effectiveStartBlock`, so server-side address filtering is redundant for client-filtered contracts.
- New dynamic addresses for wildcard contracts get a server-side backfill query up to the wildcard frontier instead of a standing partition, avoiding partition proliferation.
- Partition absorption drags the wildcard frontier back slightly; re-fetching the overlap is deduped and serves as backfill for fresh addresses.

https://claude.ai/code/session_01EvNrjtdSo2EMr34yPsyyL4

## Summary by CodeRabbit

* **New Features**
  * Added client-side address filtering for high-cardinality contract registrations, switching them to address-free (“wildcard”) routing.
  * Introduced `clientFilterAddressThreshold` to control when contracts switch modes.
  * Added `clientFilteredContracts` / `clientSideFilteredContracts` query and pagination options to carry selected contract filtering behavior through requests.
* **Bug Fixes**
  * Improved routing/decoding and partition collapse during address-free transitions to match wildcard behavior correctly.
  * Prevented duplicate event processing when contracts switch filtering modes mid-fetch.
* **Tests**
  * Expanded end-to-end and unit coverage for threshold switching, routing, rollbacks, partition behavior, and deduplication.


DZakh:
Pushed `77cbd03` addressing the review threads:

- **Shared `clientFilteredContracts` mutation** (coderabbit/codex) — `registerDynamicContracts` now clones the sticky set before adding, so older `fetchState` snapshots keep their own copy.
- **`restart()` dropping overrides** (coderabbit) — the recursive `make` in `MockIndexer.restart` now forwards `maxAddrInPartition` and `clientFilterAddressThreshold`.
- **Weak dedup-test assertion** (coderabbit) — `ClientFilterDedup_test` now snapshots the pre-switch queries and requires a *newly appended* post-switch query to cover block 10, so the initial fetch can't satisfy the overlap check on its own.

Two I'm not changing, with reasons:

- **Preserve in-flight reservations on collapse** (codex, outdated) — addressed by the collapse rewrite in #1485: the standing address-free partition keeps its id/frontier/in-flight queries, and absorbed-below-frontier partitions become a single `mergeBlock`-bounded backfill (capped by `maxInFlightChunksPerPartition`), so the concurrency isn't unbounded.
- **Gate by source support for `SimulateSource`** (codex, outdated) — the switch is gated on the EVM ecosystem, and the default threshold (`maxAddrInPartition * maxChainConcurrency / 2` ≈ 250k) is far beyond what any `indexer.process()` simulate run registers, so it can't trigger in practice. Worth a follow-up if simulate ever needs it, but leaving `SimulateSource` routing untouched for now.

Separately: the red `hypersync-health-check` job is an external DNS failure reaching `728126428.hypersync.xyz` ("Name or service not known") — it pings live HyperSync endpoints and is unrelated to this diff, which doesn't touch the endpoint list or the health test. It should clear on a re-run once DNS resolves.

---

PR #1485, opened 2026-07-24
Preserve wildcard partition state and improve address threshold logic

DZakh:
## Summary
Improves wildcard (client-side) address filtering by preserving partition state across unrelated contract registrations, applying the threshold at FetchState creation time, and making the chain concurrency limit configurable.

## Key Changes

- **Preserve wildcard partition identity**: When collapsing wildcard contracts, avoid rebuilding the partition if only the existing wildcard partition is absorbed and no new contracts have switched to wildcard mode. This preserves in-flight queries and learned density across unrelated registration batches.

- **Apply threshold at creation**: Switch contracts to wildcard mode at FetchState creation if they exceed the address threshold, not just during dynamic registration. This handles both config contracts with large static address lists and dynamic contracts restored from persisted state.

- **Make chain concurrency configurable**: Extract `maxChainConcurrency` to `Env.res` with a fallback of 100, allowing it to be overridden via `ENVIO_MAX_CHAIN_CONCURRENCY` environment variable. Update the default `maxContractServerSideAddresses` calculation to use this value dynamically.

- **Optimize address lookup**: Replace ReScript loop in `IndexingAddresses.get` with raw JavaScript for..in to avoid allocating a values array on every lookup (runs once per registration).

- **Simplify client-filtered check**: Remove unnecessary empty set check in `ClientFilteredContracts.applies` — the set membership check alone is sufficient.

- **Add logging**: Log when contracts switch to wildcard mode with context (chain ID, contract name, address count, threshold).

## Tests
Added three new test cases:
- Wildcard partition preservation across unrelated registrations
- Config contract threshold application at creation
- Wildcard mode persistence through rollback

https://claude.ai/code/session_016H5TKct8fzYmcoXaHpVMbC

### DIFF
--- a/packages/cli/src/client_filtered_contracts.rs
+++ b/packages/cli/src/client_filtered_contracts.rs
@@ -22,6 +22,6 @@ impl ClientFilteredContracts {
     /// wildcard. Only address-dependent registrations are affected; a genuinely
     /// wildcard registration is already address-free.
     pub fn applies(&self, contract_name: &str) -> bool {
-        !self.0.is_empty() && self.0.contains(contract_name)
+        self.0.contains(contract_name)
     }
 }
--- a/packages/envio/src/Env.res
+++ b/packages/envio/src/Env.res
@@ -9,17 +9,20 @@ let updateSyncTimeOnRestart =
 let targetBufferSize = envSafe->EnvSafe.get("ENVIO_INDEXING_MAX_BUFFER_SIZE", S.option(S.int))
 let maxAddrInPartition = envSafe->EnvSafe.get("MAX_PARTITION_SIZE", S.int, ~fallback=5_000)
 
+// Most parallel in-flight queries a single chain may have at once, across all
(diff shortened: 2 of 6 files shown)


DZakh:
The `hypersync-health-check` failure is environmental, not a code issue. The job panicked on a DNS lookup, not a test assertion:

```
Failed to fetch health for https://4200.hypersync.xyz after 3 retries:
error trying to connect: dns error: failed to lookup address information: Name or service not known
```

That check (`packages/cli/tests/hypersync_health.rs`) pings live HyperSync endpoints and is untouched by this PR. The runner couldn't resolve the chain-4200 endpoint. A re-run once the endpoint resolves should clear it — no code change is warranted here.

The checks that actually exercise this diff passed: `cargo-test` ✅ and `build-envio-package` ✅ (`e2e-test`, `template-tests`, `scenarios-test` still running at time of writing).

---

PR #1527, opened 2026-08-03
Remove redundant index catalog logging from PgStorage

DZakh:
Removes a logging statement that reports the number of existing indexes found in the indexer schema during storage initialization.

## Summary
This log message was informational but redundant — it provided visibility into an intermediate step of the initialization process without actionable value. The initialization itself succeeds or fails based on index validation, which is already reported via error logging when invalid indexes are detected.

## Changes
- Removed the `Logging.info` call that reported the count of existing indexes in the indexer schema

https://claude.ai/code/session_016F5JKQ7HHN8mdHV2XXrQQY

## Summary by CodeRabbit

* **Chores**
  * Removed an informational log message displayed while reloading PostgreSQL indexes.
  * Removed the obsolete Sophon testnet network option.

* **Documentation**
  * Updated blockchain selection guidance to list Sophon instead of Sophon testnet for contract imports.


DZakh:
`hypersync-health-check` is red for a reason unrelated to this PR: the Sophon Testnet endpoint is refusing connections.

```
Failed to fetch health for https://531050104.hypersync.xyz after 3 retries:
error trying to connect: tcp connect error: Connection refused (os error 111)
```

It passed on the first run of this PR (12:18 UTC) and failed on the second (13:21 UTC) with no code change in between. Probing directly still gets a connection reset, while Sophon mainnet (`50104.hypersync.xyz`) returns 200 — so this looks like the testnet endpoint being down or retired, not a regression here. Not fixing it in this PR; deciding whether `SophonTestnet` should stay in `HypersyncChain` is a separate call. Will re-run the job once the endpoint recovers.

---

PR #1528, opened 2026-08-03
SVM: upgrade to HyperSync Solana 0.2.0-rc.4 (Wave 2 API) and the mainnet-history endpoint

JasoonS:
Upgrades the SVM source to the HyperSync Solana Wave 2 API (crates `hypersync-client-solana` / `hypersync-solana-net-types` `0.2.0-rc.4`) and switches the default Solana endpoint to `https://solana-mainnet-history.hypersync.xyz`.

## Upstream delta (0.0.7 -> 0.2.0-rc.4)

- Renames: query key `instructions` -> `instruction_calls`; selection/response field `program_id` -> `executing_account`; instruction `accounts` -> `account_arguments`; `field_selection.instruction` -> `instruction_call`; `is_committed` -> `tx_success` (now also a server-side instruction-selection filter).
- The `balances` and `token_balances` tables are removed, replaced by the unified `account_activity` table (native + token sides merged, `owner` split into `pre_owner`/`post_owner`, token amounts as u64 with string serde).
- `#[serde(deny_unknown_fields)]` on the query envelope: removed keys (`balances`, `token_balances`, `include_account_activity`) now fail the whole query, so nothing legacy survives in what we serialize.
- Response structs are all-`Option` with base58 byte newtypes (`Address`/`Hash`/`Signature`).

## What maps where

- **tokenBalances -> account_activity**: the public `svmTransaction.tokenBalances` field is now sourced from token-side `account_activity` rows (rows carrying a `mint`), requested purely via `field_selection.account_activity` (the join is keyed off field selection; `include_account_activity` no longer exists). `owner` maps from `post_owner`, falling back to `pre_owner` for accounts closed in-transaction; `preAmount`/`postAmount` come from the u64 token balances rendered as decimal strings. Public handler types (`svmInstruction`, `svmTransaction`, `svmTokenBalance` in `Envio.res` / `index.d.ts`) are unchanged.
- **Owned row structs at the napi boundary**: rc.4's all-`Option` newtype responses are converted once into `SvmBlockRow` / `SvmTxRow` / `SvmTokenBalanceRow`, so the block/transaction stores keep borrowing plain strings.
- **Failed-transaction exclusion moves server-side**: every instruction selection HyperIndex builds now carries `tx_success: true`. This supersedes the unmerged `js/svm-exclude-failed-tx` branch (client-side `isCommitted` filter); a cheap client-side guard on the response's `txSuccess` flag remains as defense-in-depth, with a mocked-client test.
- **Endpoint**: every resolution of the Solana mainnet endpoint now points at `solana-mainnet-history.hypersync.xyz`. Queries on the new endpoint require a bearer token (`ENVIO_API_TOKEN`); the skipped/ignored live tests read it from the environment.

## Verification

- `cargo check` clean; `cargo test --lib`: 410 passed, 1 failed (`executor::tools::tests::search_docs_clickhouse_support`, a live docs-MCP snapshot test whose upstream content drifted; unrelated to this change).
- `pnpm rescript` clean in `packages/envio` and `scenarios/test_codegen`; `tsc --noEmit` clean (public TS surface unchanged).
- Vitest: `SvmHyperSyncSource_test` (incl. new failed-tx drop test), `EventRouter_svm_test`, `BlockStore_test`, `TransactionStore_test`, `svm_test` scenario all pass.
- Live smoke against the new endpoint:
  - `/height` ~436.45M; raw `POST /query` with `instruction_calls`/`executing_account`/`tx_success` returns the new wire shape.
  - Rust ignored live test (`live_query_token_metadata`) and the ReScript live client test both pass through the real napi client.
  - `account_activity` join semantics confirmed: with only `field_selection.account_activity` set, returned rows are exactly those keyed to transactions matched by the instruction selections.
  - Failed-tx exclusion confirmed: in a sampled 20-slot window the transactions table serves 852 failed transactions, but zero instruction rows exist for them (the endpoint ingests with the failed-transaction trim), and filtered vs unfiltered instruction queries return identical `tx_success=true`-only row sets. Our `tx_success: true` filter plus the client guard make the exclusion a guarantee rather than an ingest-policy accident.
  - End-to-end: `svm_flow_xray` and `svm_metaplex_demo` live E2E indexer runs pass (instructions, CPI nesting, token deltas via account_activity, entity writes).

## Notes for reviewers

- The new endpoint's history currently begins at slot 422,000,000; the pinned E2E windows moved accordingly. Slots below the served range return empty pages with an advancing `next_slot` (no error), which is worth knowing when repinning windows.
- The `reward` table plumbing is kept (still present upstream); the napi response no longer carries a `balances` array (nothing consumed it).

## Refactor pass

Follow-up commits on the same branch, no behavior change (re-verified live):

- **Endpoint as a single source of truth**: the Solana mainnet URL now lives in `SOLANA_MAINNET_HYPERSYNC_URL` (`packages/cli/src/config_parsing/hypersync_endpoints.rs`), next to the EVM endpoint resolution. `experimental.hypersync_config` is now optional and defaults to it, matching the EVM convention; the metaplex template relies on the default. Covered by a new system_config test; `svm.schema.json` regenerated. Scenario/test fixture YAMLs keep literal URLs by design.
- **Module layout**: `borsh_decoder.rs` renamed to `decode.rs` so the SVM source directory reads like the EVM one (config / decode / query / types / mod).
- **Type tightening**: `SvmTokenBalanceRow.mint` is now a plain `String` (a token row without a mint was representable but never valid); the post-owner-falling-back-to-pre rule is a named function (`owner_at_transaction_end`).
- **ReScript alignment with EVM/Fuel**: the SVM source now requires the Envio API token at construction with the same fail-fast message as the EVM and Fuel sources (the endpoint rejects unauthenticated queries, so a missing token previously surfaced as opaque 401 retries); client instantiation is wrapped in the same `mkLogAndRaise` shape; the instruction selections and query field selection are assembled once at `make` time instead of per fetch; the wire-name field enums are completed against the rc.4 tables.

Verification after the pass: `cargo check` clean, `cargo test --lib` 411 passed (same single pre-existing unrelated live-docs snapshot failure), `pnpm rescript` clean, SVM vitest files pass, and the live Rust smoke test plus the `svm_flow_xray` live E2E both pass against the endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added unified Solana account activity queries with filters for accounts, mints, owners, activity types, and programs.
  * Added instruction-call queries with executing-account and transaction-success filters.
  * Added support for authenticated Solana mainnet history access.

* **Bug Fixes**
  * Failed-transaction instructions are now excluded from processing.
  * Improved validation for malformed public keys and response data.

* **Documentation**
  * Updated Solana configuration examples and setup instructions to use the mainnet history endpoint.


JasoonS:
Stale

PR #1539, opened 2026-08-11
Add inline field selection for EVM event handlers

DZakh:
## Summary

Adds an optional `fields` parameter to `indexer.onEvent()` and `indexer.contractRegister()` for EVM chains, allowing handlers to declare which block and transaction fields they read. This replaces the global `field_selection` from `config.yaml` on a per-registration basis, enabling different handlers on the same event to select different fields while the indexer fetches their union.

Along the way it corrects the `field_selection` validation, which rejected 35 fields on RPC-synced chains that RPC endpoints do return — see [RPC field availability](#rpc-field-availability) below.

## Key Changes

- **Type system** (`index.d.ts`): Added `EvmFieldsSelection`, `EvmAllBlockFields`, `EvmAllTransactionFields` types and `EvmEventWithFields` conditional type that narrows event block/transaction fields based on the inline selection. Unselected fields become `FieldNotSelected` to produce helpful compile errors. A selection that isn't a literal array is rejected (`EvmFieldsMustBeLiteral`): read from a variable typed `EvmFieldsSelection`, the element types name the whole field union rather than the selection, which would type every field as selected while the runtime selected a subset.

- **Handler registration** (`HandlerRegister.res`):
  - Added `fields` parameter to `buildOnEventRegistrationWith`
  - Implemented `unionSelection` to merge field selections when a `contractRegister` merges into a handler (or vice versa)
  - Updated `mergeRegistrations` to union block/transaction field masks when combining registrations
  - Added RPC validation to reject fields unavailable via RPC when a chain syncs over RPC. It runs on the registrations a chain keeps, so a handler whose `where` opts out of an RPC chain isn't held to its limits, and its message names neither `fields` nor `field_selection` since either can be the source.

- **Field resolution** (`EventConfigBuilder.res`):
  - Implemented `parseFieldsOrThrow` to validate field names, reject unrecognised keys and detect duplicates
  - Implemented `resolveRegistrationFieldSelection` to apply inline `fields` or fall back to event config selection
  - Inline selections always include `block.number` without listing; when `raw_events: true`, also include `block.hash` and `block.timestamp` for `toRawEvent` materialisation

- **Runtime** (`Internal.res`): Added a shared `fieldSelection` to `onEventRegistration` so each registration carries its own selection and masks independent of the shared `eventConfig`.

- **Data fetching** (`RpcSource.res`, `HyperSyncClient.res`): Updated to read field selections from the registration rather than the event config.

- **Tests** (`EvmInlineFieldSelection_test.res`, `RpcFieldSelection_test.res`): Selection replacement, fallback to config, always-included fields, handler separation, registration merging, raw events, RPC validation and error cases; plus a bidirectional drift guard tying the config-parse validation to the runtime's own field registry.

- **Documentation**: Updated skill guides and handler examples to show the new `fields` option.

- `block.number` is always selected for inline registrations (it's the item's own key, so it costs nothing extra). When `raw_events: true`, `block.hash` and `block.timestamp` are also always selected because `toRawEvent` reads them from the payload.
- Two registrations of the same event can select different fields; the indexer materialises the union so both can dispatch. A field one handler listed is therefore often populated for another that didn't — the types, not the runtime, are what hold each handler to its own selection.
- Inline `fields` completely replaces the config `field_selection` for that registration — a field selected in config but not listed is a compile error in that handler.
- A test in `codegen_templates.rs` ensures the hand-written `EvmAllBlockFields`/`EvmAllTransactionFields` types in `index.d.ts` stay in sync with the config-parsing enums.

## RPC field availability

`field_selection` refused 20 transaction fields and 15 block fields whenever a chain had an RPC sync source. Live probing shows RPC endpoints return them, so all 35 are now selectable:

**Transaction** (20) — `gas`, `nonce`, `v`, `r`, `s`, `yParity`, `type`, `maxFeePerBlobGas`, `blobVersionedHashes`, `cumulativeGasUsed`, `effectiveGasPrice`, `gasUsed`, `logsBloom`, `root`, `status`, `l1Fee`, `l1GasPrice`, `l1GasUsed`, `l1FeeScalar`, `gasUsedForL1`

**Block** (15) — `sha3Uncles`, `logsBloom`, `transactionsRoot`, `receiptsRoot`, `totalDifficulty`, `size`, `uncles`, `blobGasUsed`, `excessBlobGas`, `parentBeaconBlockRoot`, `withdrawalsRoot`, `l1BlockNumber`, `sendCount`, `sendRoot`, `mixHash`

Still rejected: `accessList` and `authorizationList`. The RPC does return them, but their array shape has no parser in the runtime's field registry, so they would be silently dropped.

### Why this was possible

Nothing changed in what the RPC source can do — the capability was already there and already under test:

- `RpcSource.res`'s `makeFieldRegistry` has a parser for every transaction field except those two, split across `eth_getTransactionByHash` (`gas`, `nonce`, `v`/`r`/`s`, `yParity`, blob fields), `eth_getTransactionReceipt` (`gasUsed`, `cumulativeGasUsed`, `effectiveGasPrice`, `logsBloom`, `root`, `status`, the `l1*` family) and both (`type`). `makeBlockFieldRegistry` covers every block field.
- `RpcSource_test.res` already asserted `blockFieldRegistry contains all evmBlockField variants` — the runtime guaranteed complete block coverage while the config refused 15 of 24 — and its real-RPC test already asserted `gas`, `nonce`, `value`, `v`/`r`/`s`, `yParity`, `gasUsed`, `effectiveGasPrice` and `status` parse from a live response, all of which the config rejected.

The gate was the `RpcTransactionField`/`RpcBlockField` allowlist in the config parser, which never tracked those parsers. No consistent rule separated what it allowed from what it refused: `value` was allowed but `nonce` wasn't, `gasPrice` but not `gas`, and `contractAddress` was allowed while every other receipt-sourced field was refused — so it wasn't "receipts aren't fetched", just staleness.

Probing four chains confirmed the endpoints return them: mainnet for the block fields and the core transaction fields, BSC for `totalDifficulty` (post-merge chains drop it), Arbitrum for `l1BlockNumber`/`sendCount`/`sendRoot`, Optimism for the `l1Fee` family. Fields that didn't come back are chain- or transaction-specific rather than RPC limitations — `root` is pre-Byzantium, the blob fields need a type-3 transaction — and the runtime already decodes those as nullable.

`RpcFieldSelection_test.res` now holds the config validation to the runtime registry in both directions, and `RpcSource_test.res` pins the block fields with no HyperSync equivalent against a real RPC response.

One field did need work to admit: pre-Bedrock Optimism receipts omit `effectiveGasPrice` entirely, which crashed the source until #1546 made it nullable.

https://claude.ai/code/session_011usC7i2cdPhW6WwdZrEXCr

## Summary by CodeRabbit

* **New Features**
  * Added inline EVM block and transaction field selection for event and contract registrations.
  * Handlers now receive type-safe event data, with compile-time errors for unselected fields.
  * Added validation for unsupported and duplicate fields, required-field inclusion, and per-registration overrides.
  * Expanded RPC support for additional transaction and block data.

* **Documentation**
  * Expanded guidance for inline field selection and transaction/block data access.

* **Tests**
  * Added comprehensive coverage for selection, validation, merging, RPC support, and generated declarations.


packages/cli/templates/static/shared/.claude/skills/indexer-transactions/SKILL.md
```diff
@@ -1,63 +1,72 @@
...
-Or globally for all events:
+Reading a field you didn't list is a compile error, so the selection and the
+handler can't drift. `block.number` is always readable; everything else has to
+be listed.
+
+Two handlers on one event can select different sets. Read only what you listed —
+anything else is a type error, and its value is not guaranteed.
+
+Write the selection inline, or in a variable declared `as const`. A variable
+typed `EvmFieldsSelection` is rejected, because it no longer says which fields
```
DZakh: I don't understand this line

PR #1544, opened 2026-08-13
Migrate scenario tests to new IndexerRunner infrastructure

DZakh:
Consolidates test infrastructure by moving scenario test helpers from the generated `test_codegen` project into the shared `envio-tests` package, enabling tests to run against multiple storage backends (memory, Postgres, ClickHouse) without code duplication.

## Summary

This refactors the test architecture to separate concerns: test machinery (sources, storage mocks, runner) lives in `envio-tests` as reusable infrastructure, while individual test scenarios live in `packages/envio-tests/test/` and can declare which backends they support. The generated `test_codegen` project now only hosts legacy tests still using the old `MockIndexer` API.

## Key Changes

- **New `IndexerRunner` module** (`packages/envio-tests/test/helpers/IndexerRunner.res`): Unified test runner that executes scenarios against selected backend (memory, Postgres, or ClickHouse via `ENVIO_TEST_STORAGE` env var). Handles schema isolation, client lifecycle, and provides query/metric inspection APIs.

- **New `MemoryStorage` module** (`packages/envio/src/MemoryStorage.res`): In-process storage implementation with full entity history, rollback support, and effect cache — enabling all test scenarios to run without a database.

- **New `Scenario` module** (`packages/envio-tests/test/helpers/Scenario.res`): Declarative test API where scenarios specify config YAML, schema, and handlers. Supports marking backends as unsupported with explanations.

- **Extracted `MockSource` and `MockStorage`** (`packages/envio-tests/test/helpers/`): Moved from `MockIndexer` to shared infrastructure. `MockSource` provides event injection with per-test handler dispatch; `MockStorage` mocks the persistence layer for legacy tests.

- **Migrated test files**: Moved 8 scenario tests from `scenarios/test_codegen/test/` to `packages/envio-tests/test/`:
  - `BlockLag_test.res`, `BelowHeadPollingPin_test.res`, `ClientFilterDedup_test.res`
  - `HeightPushMetric_test.res`, `MultichainStuckAtHead_test.res`, `SchemaIsolation_test.res`
  - `SparseDensityTargetOverflow_test.res`, `StalledPolling_test.res`
  - Plus new tests: `ScenarioRollback_test.res`, `ScenarioIsolation_test.res`, `ScenarioClickHouse_test.res`

- **Updated `MockIndexer`** (`scenarios/test_codegen/test/helpers/MockIndexer.res`): Now a thin surface re-attaching generated `Indexer` types to the shared machinery. Marked as legacy with a comment explaining it goes away when the last test migrates.

- **ClickHouse support**: Added `TestClickHouse.res` to manage per-test ClickHouse databases, and `.claude/hooks/start-clickhouse.sh` to start the service for local development.

- **CI integration**: Updated `vitest.config.ts` in both packages to run tests in parallel with proper isolation. Modified GitHub Actions workflow to run `envio-tests` once per backend.

## Implementation Details

- **Backend selection**: Tests run against the backend specified by `ENVIO_TEST_STORAGE` env var (defaults to `#memory`). Scenarios can declare unsupported backends via `~unsupported` parameter.

- **Isolation**: Each test run gets its own Postgres schema (via `TestPgSchema`) or ClickHouse database, allowing parallel execution against a single server.

- **Memory persistence**: `MemoryStorage` keeps entities decoded in-process, supporting real bigint/BigDecimal values and full history tracking — matching Postgres semantics without a database.

- **Handler registration**: `HandlerRegister.makeRegistration` creates isolated registrations per test, preventing cross-contamination when multiple scenarios run in one process.

- **User module loading**: `UserModule.res` writes handler source to disk and imports it as a real module, exercising the actual module boundary instead of hand-built registration.

https://claude.ai/code/session_01YZ64Nodb93aSDzyJ13x6jm

## Summary by CodeRabbit

* **New Features**
  * Added scenario testing across in-memory, PostgreSQL, and ClickHouse storage.
  * Added automated ClickHouse setup with health checks and readiness handling.
  * Improved support for isolated scenarios, schemas, and handler registrations.

* **Bug Fixes**
  * Addressed polling stalls, lagged chains, rollback recovery, duplicate event processing, and dynamic queue behavior.
  * Improved concurrent writes, history persistence, and multichain synchronization.

* **Testing**
  * Expanded coverage for reorganization thresholds, ClickHouse persistence, dynamic contracts, and cross-backend behavior.
  * Consolidated test helpers and enabled concurrent test-file execution.


DZakh:
Addressed the review in b19774a. Verified on all three backends (memory / postgres / clickhouse) before pushing.

**Fixed**

- **Unmocked chains** — real bug, thanks. `withMockSources` left an unmocked chain's `sourceConfig` at whatever the YAML named, and the runner starts every configured chain, so it would have queried a live URL. It now requires the mocked and configured chain sets to match exactly, and rejects mocks for chains the YAML doesn't declare.
- **`enterReorgThreshold` yield** — it asserted before yielding and only worked because every current caller happened to yield first. Now yields itself.
- **`DynamicContractPersistence` handlers** — added, though not for the stated reason (see below).
- Bounded `ConcurrentWrite`'s two waits, annotated the `Obj.magic` cast per the repo's ReScript guidelines, broke the `DynamicSplitQueueAlias` sort tie on partition id, trimmed the module-purpose narration in `Scenario.res`, and corrected the `max_reorg_depth` comment in `EnterReorgThreshold`.

**One correction**

> Without `~handlers`, the `contractRegister` callbacks are never registered, so the test does not create dynamic partitions or exercise persistence.

Not accurate as stated. `MockSource` installs a synthetic registration carrying the item's `contractRegister`, so the callbacks did run — the test asserts exact `envio_addresses` rows (`[("1337-0x1111…", "Gravatar", 10)]`), which only appear if registration happened, and the third case asserts the drain splitting across batches, which requires the partition to spawn.

What's true is narrower: a dynamically registered address only spawns a partition when some registration depends on its contract, and that held only because `Gravatar` happens to be `contracts[0]`, which is what the synthetic registration binds to. That's fragile, and `NftFactory` had no dependent registration — so the conflict case could have passed for the wrong reason. Adding `~handlers` removes both, so I made the change.

**Not changed**

The mergeability note says parallel ClickHouse tests may interfere. Each run gets its own database (`TestClickHouse.make`, dropped in teardown), tests within a file are sequential (`sequence.concurrent: false`), and vitest forks give each file its own process env — so there's no shared state to collide on. Happy to reconsider with a concrete interleaving.

---

DZakh:
Second round addressed in d10d979 — all three backends re-run green (451 passed / 4 skipped on memory and postgres, 451 / 4 on clickhouse).

**Fixed**

- **ClickHouse env race** — good catch on the mechanism. `TestClickHouse.make()` set `ENVIO_CLICKHOUSE_DATABASE` at the top of `run`, and `await resolveRegistrations()` sat between that and the read in `makeStorageFromEnv`. No current test overlaps two runs (tests are sequential within a file, and each `Scenario.run` is awaited), so it isn't reachable today — but the window shouldn't exist. `make()` now only mints the name; a separate `use(~database)` sets the env immediately before the storage is built, with no await in between.
- **`MultichainStuckAtHead` timeout** — my regression: the original carried `~timeout=60_000` and `Scenario.it` had no way to express it, so it silently fell back to the 30s default. `Scenario.it` now takes `~timeout` and the test asks for 60s again.
- **`StalledPolling`** — collapsed to a single assertion with a message, and widened the upper bound to 15; the old `<= 8` was tight enough to flake under timer coalescing.
- **`SchemaIsolation`** annotated cast, **`BlockLag`** stale `driveToFetch` reference removed.

**Not changed — `SiblingRollbackStuckChain` catch-up tracking**

Tracking `latest` instead of `fromBlock` would weaken what this test pins. The incident it guards is *"getNextQuery never produced a query again"*: after the rollback the victim has to fetch its rolled-back range **and then ask for the next block**. Requiring a query that *starts* at 302 proves it issued a query from the head; requiring a response that *reaches* 302 does not — a chain that resolves one query to 302 and then never queries again would pass the proposed version and fail the current one.

The hypothetical failure (a final 301→302 request leaving the counter at 301) isn't reached: the loop keeps answering height polls at 302, so the chain issues the 302 query once its frontier gets there. The logic is also inherited verbatim from the pre-existing test and passes on all three backends. Happy to revisit if it ever flakes in CI.

---

DZakh:
⚠️ **The new CI matrix in this PR does not run on this PR.**

`Build & Verify` triggers on `pull_request_target`, so GitHub takes the workflow *definition* from `main`. The three `envio-tests (memory|postgres|clickhouse)` steps and the ClickHouse service container added here are inert until merge — `main`'s single `envio-tests` step runs instead, with no `ENVIO_TEST_STORAGE`, i.e. **memory only**.

The job does check out this branch's code (`ref: pull_request.head.sha`), so the migrated tests are genuinely exercised — just on one backend.

Postgres and ClickHouse are verified locally: every migrated file was run on all three legs before its original was deleted, and the full suite is green on each (454 / 4 skipped on postgres, 453 / 5 on clickhouse, 451 / 7 on memory).

Supporting evidence for the ClickHouse leg post-merge: the existing `e2e-test` job already runs the same `clickhouse/clickhouse-server:26.2.15.4` service and reaches it on `localhost:8123`, which is what `TestClickHouse` defaults to.

Worth a reviewer's eye on the workflow diff specifically, since CI can't check it here.

---

DZakh:
✅ **The backend matrix is now verified in CI** — [run 31715676187](https://github.com/enviodev/hyperindex/actions/runs/31715676187/job/94500295828).

Resolves the caveat above. A `push` trigger for this branch was added temporarily (64c1367), because `push` events use the workflow definition from the branch rather than from `main`. `scenarios-test` ran all three legs against the real service containers:

| step | result |
|---|---|
| `envio-tests (memory)` | 450 passed, 7 skipped |
| `envio-tests (postgres)` | 453 passed, 4 skipped |
| `envio-tests (clickhouse)` | 452 passed, 5 skipped |

Identical to the local numbers, and the ClickHouse service container is in the job's container log. `test_codegen` (675), `fuel_test`, `svm_test` and `cross_chain_test` all green in the same job.

Also in this push: `origin/main` merged in. The Rust `BlockStore` reorg work (#1405) conflicted with the extracted `MockSource` — responses now carry a block store instead of a `blockHashes` array, hashes widen to 32 bytes, and the source records an `onReorg` count. Ported, and the migrated rollback tests moved to the new hash markers.

⚠️ **The temporary `push` trigger must be removed before merge.** It's the only entry under `on.push.branches` besides `main` and is commented as temporary. I've left it in place while the remaining files are migrated so each push keeps getting matrix coverage — say the word and I'll drop it.

---

DZakh:
Ready to merge.

- **`main` merged in** (85ea1b8). #1539 collapsed `selectedBlockFields` / `selectedTransactionFields` / the two masks into a single `fieldSelection`, which conflicted with the mock registration extracted into `MockSource`. Ported there, plus the new `fieldSelection` on `onEventRegistration`. Base is now `0289eb5`; `mergeable_state: clean`.
- **Temporary CI trigger reverted** (edf7f01). `on.push.branches` is back to `main` alone; the workflow diff against main is now only the ClickHouse service and the three `envio-tests` steps.

Verified after the merge — all three legs locally (memory 478/7, postgres 481/4, clickhouse 480/5) and test_codegen 678. CI green on edf7f01.

Note the matrix evidence remains [run 31715676187](https://github.com/enviodev/hyperindex/actions/runs/31715676187/job/94500295828); this PR's own run exercises the memory leg only, since `pull_request_target` takes the workflow from `main`. The matrix starts running for real on the merge commit.

---

PR #1568, opened 2026-08-20
Add @internal entity directive to hide entities from the GraphQL API

DenhamPreen:
## What

Adds a no-arg `@internal` directive on entities in `schema.graphql`. An internal entity is stored in Postgres/ClickHouse per its `@storage` resolution and fully usable in handlers (`context.X.set/get`, generated types, test helpers), but is never exposed through the GraphQL API — not tracked in Hasura, no select permission, no relationships. Composes freely with `@crossChain` and `@storage`.

```graphql
type Secret @internal {
  id: ID!
  note: String!
}
```

Motivation: users want entities for internal indexing state (or sensitive data) without exposing them publicly through GraphQL.

## How

- **Parsing** (`entity_parsing.rs`): `internal: bool` on `Entity`, parsed like `@crossChain` — rejects arguments and duplicate directives.
- **Validation** (`system_config.rs`): new `validate_internal_relationships` fails codegen when an exposed entity references an `@internal` one (object reference or `@derivedFrom`), since Hasura can't create a relationship to an untracked table and a derived field would silently vanish from the API. Internal→public references stay legal.
- **Persisted config** (`public_config.rs`): `"internal": true` emitted only when set, so every existing project's persisted config stays byte-identical (no snapshot diffs, no compat-check trips on resume).
- **Runtime** (`Internal.res`, `Config.res`, `Hasura.res`): `entityConfig` carries the flag; tracking-table construction is extracted into a pure `Hasura.makeTableConfigs` that filters internal entities (permissions follow automatically), and the relationship loops iterate only exposed entities. `getPgUserEntities` is untouched, so storage, loaders, and history tables are unaffected.

## Tests

- Rust: directive parse tests (flag, args rejected, duplicates rejected), relationship-validation tests (both error directions plus allowed internal references), and a config-JSON shape test asserting `internal` is present only when set.
- `packages/envio-tests/test/InternalEntity_test.res`: full user API — real YAML + schema + TS handlers proving a handler sets/reads an `@internal` entity while `Hasura.makeTableConfigs` excludes it.
- e2e: `TransferInternal @internal` added to `scenarios/e2e_test` with assertions that its Postgres table exists with full row counts while the introspected GraphQL schema never mentions it.

## Notes

- Toggling `@internal` on a deployed indexer trips the persisted-config compat check like any other entity-level change (same as `@crossChain`/`@storage` edits today).
- Docs follow-up: `@internal` needs a reference entry in the docs repo alongside `@crossChain`/`@storage`.

## Summary by CodeRabbit

* **New Features**
  * Added support for `@internal` entities that remain stored and accessible to handlers without exposure through the GraphQL API.
  * Internal entities are reflected in generated configuration and excluded from public table tracking and relationships.
  * Added validation preventing exposed entities from referencing internal entities.

* **Bug Fixes**
  * Added validation for invalid or duplicate `@internal` directives.
  * Improved internal entity configuration and storage coverage through end-to-end tests.

* **Improvements**
  * Removed deprecated per-instruction transaction and block fields from generated SVM types.


DenhamPreen:
This will have implications on envio serve too

DenhamPreen:
Re: `envio serve` — confirmed. serve builds its GraphQL surface by parsing `schema.graphql` directly (`packages/cli/src/serve/project_schema.rs` on the serve branch collects every object type), so an `@internal` entity's table would be exposed there even though Hasura skips it. Since serve isn't on main yet, the filter belongs on that branch when it rebases; tracked in #1569 with the specifics (skip `@internal` entities in `ProjectSchema::parse`, drop relationships pointing at them, and add a differential-corpus case).

DZakh:
This is a good change. Let me know when it's ready and ci passes and I'll take a look at the code.

packages/cli/src/config_parsing/system_config.rs
```diff
@@ -843,6 +843,44 @@ pub fn validate_cross_chain_relationships(
     ))
 }
 
+/// An @internal entity has no GraphQL surface, so a relationship from an
+/// exposed entity to it cannot be served: an object reference would break
+/// Hasura relationship creation against the untracked table, and a
+/// @derivedFrom field would silently vanish from the API. Reject both at
+/// codegen. References from @internal entities are fine — nothing is exposed
+/// on their side.
```
DZakh, line 851: I think it would be nice to allow this case, just not create a relationship if it targets the internal entity

DZakh (approved):
Looks good, left one comment but it can be easily done as a follow-up when it's requested by someone.

PR #1576, opened 2026-08-21
SVM IDL catalog: parse Anchor/Codama, drop YAML instruction lists

DZakh:
## Summary

SVM programs are configured with `name`, `program_id`, and `idl`. The IDL is the catalog for instruction names, discriminators, and Borsh layouts. YAML `instructions:` is rejected.

Register `indexer.onInstruction({ program, instruction, fields })` for the instructions to handle. Payload selection stays on the handler `fields` option (not YAML `field_selection`).

Config errors say what to do in place. They no longer point at the docs.

## Config

```yaml
experimental:
  programs:
    - name: TokenMetadata
      program_id: metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s
      idl: idls/token-metadata.codama.json
```

- `idl` is required (Anchor or Codama).
- Unusable IDL instructions are omitted; if none remain, the program fails with reasons.
- Two programs may not share a `program_id`.

## Also in this branch

- In-tree Anchor (legacy + 0.30) and Codama parsers.
- Prefix-collision dispatch (1/2/4/8-byte, longest first), including set-aside instructions.
- Merged with #1559: keep name-keyed IDL resolution, drop YAML `field_selection`.
- Raydium / Orca / Meteora in flow-xray use small stand-in Anchor IDLs (not real Codama conversions).

## Summary by CodeRabbit

* **New Features**
  * Added support for Anchor and Codama IDLs, including instruction, account, and type discovery.
  * Solana mainnet and devnet now use public HyperSync endpoints by default.
  * SVM handlers now provide stronger type-safe field selection and nested account data.
  * Added IDL-based configuration for Metaplex and Flow X-Ray examples.
* **Bug Fixes**
  * Improved validation and diagnostics for invalid IDLs, duplicate accounts, conflicting discriminators, and unsupported layouts.
  * Improved SVM simulation payloads and missing-instruction guidance.
* **Documentation**
  * Updated SVM and Metaplex configuration guidance for IDL-based setup.


packages/cli/src/config_parsing/svm_idl/anchor.rs
```diff
@@ -0,0 +1,291 @@
...
+        if a.get("accounts").is_some() {
+            out.extend(parse_accounts(a.get("accounts"))?);
+            continue;
+        }
+        out.push(IdlAccount {
+            name: required_str(a, "name")?.to_string(),
+            optional: flag(a, "optional", "isOptional"),
+            writable: flag(a, "writable", "isMut"),
+            signer: flag(a, "signer", "isSigner"),
+        });
```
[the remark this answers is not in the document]
DZakh: Addressed in 10da4ddcaa0df0716d6f2b2712f2f251fcf74fef: `to_instruction_schema` now copies `writable` and `signer` from `IdlAccount` onto `SvmNamedAccount` instead of forcing both to false. Decoder still treats accounts as optional on the wire.

packages/cli/src/config_parsing/svm_idl/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+        FieldType::Struct(fields) => fields
+            .iter()
+            .try_for_each(|f| references_resolve(&f.ty, idl, bad)),
+        FieldType::Enum(variants) => variants
+            .iter()
+            .flat_map(|v| v.fields.iter().flatten())
+            .try_for_each(|f| references_resolve(&f.ty, idl, bad)),
+        _ => Ok(()),
+    }
+}
```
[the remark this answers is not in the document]
DZakh: Addressed in 10da4ddcaa0df0716d6f2b2712f2f251fcf74fef: validation demotes `Defined` cycles that never pass through `option`/`vec` into `unusable_types` (and any instruction that reaches them). Option-terminated graphs still parse. Shared DAGs stay O(n) via a seen set.

packages/cli/src/cli_args/interactive_init/fuel_prompts.rs
```diff
@@ -107,7 +107,7 @@ impl Contract for SelectedContract {
     }
 
     fn add_network(&mut self) -> Result<()> {
-        todo!("Fuel supports only one network at the moment")
+        Err(anyhow!("Fuel indexers support a single network"))
```
DZakh: This is only for the init flow. Right now the error is misleading.

packages/cli/src/cli_args/init_config.rs
```diff
@@ -337,7 +337,7 @@ pub mod svm {
 
     #[derive(Clone, Debug, ValueEnum, Serialize, Deserialize, EnumIter, EnumString, Display)]
     pub enum Template {
-        #[strum(serialize = "Metaplex Token Metadata (instructions) (Experimental)")]
+        #[strum(serialize = "Metaplex Token Metadata (Experimental)")]
```
DZakh:
```suggestion
        #[strum(serialize = "Experimental: Metaplex Token Metadata (onInstruction)")]
```

packages/cli/src/config_parsing/svm_idl/tests/snapshots/envio__config_parsing__svm_idl__tests__parses_legacy_anchor_jupiter_idl.snap
```diff
@@ -0,0 +1,55 @@
+---
+source: packages/cli/src/config_parsing/svm_idl/tests.rs
+expression: "render(&parse_fixture(\"jupiter\"))"
+---
+address: -
+instruction aldrinSwap 0xfbe877a6e1b9a9a1 (swapProgram, pool, poolSigner, poolMint:w, baseTokenVault:w, quoteTokenVault:w, feePoolTokenAccount:w, walletAuthority, userBaseTokenAccount:w, userQuoteTokenAccount:w, tokenProgram) ()
```
DZakh: Why do we have snapshots for IDL files ?

packages/cli/templates/static/svm_metaplex_template/typescript/schema.graphql
```diff
@@ -2,6 +2,9 @@ type TokenMetadataAccount {
   id: ID!
   mint: String!
   updateAuthority: String
+  name: String
+  symbol: String
+  uri: String
```
DZakh: Why is the change and why are the fields optional?

packages/cli/src/cli_args/interactive_init/fuel_prompts.rs
DZakh: Fixed in 023b890ae. `add_network` is only on the init `Contract` trait, and Fuel never offers that option (`can_add_network: false`). The error now says so: "The Fuel init flow does not support adding another network".

packages/cli/src/cli_args/init_config.rs
DZakh: Applied in 023b890ae: `Experimental: Metaplex Token Metadata (onInstruction)`.

packages/cli/src/config_parsing/svm_idl/tests/snapshots/envio__config_parsing__svm_idl__tests__parses_legacy_anchor_jupiter_idl.snap
DZakh: They were dumping the full parsed catalog of third-party IDLs. That is extra: layout and discriminator rules already have targeted tests. 023b890ae drops the insta files. Jupiter/Drift/Kamino now pin `(address, instruction count, unusable count)`. SPL Token and Memo pin the demotion reasons those fixtures exist for.

packages/cli/templates/static/svm_metaplex_template/typescript/schema.graphql
DZakh: Extra. The catalog work only needed `idl:` and IDL account names (`updateAuthority`). `name`/`symbol`/`uri` were a template expansion that made the fields optional because Update can omit `data`. Reverted in 023b890ae: schema and handler are accounts-only again.

PR #1577, opened 2026-08-21
Add transaction fields to SVM simulate items

DZakh:
## Summary

Extends SVM simulate items to carry transaction metadata (signature, fee payer, success status, compute units, etc.) through the indexing pipeline, making this data available to handlers via `instruction.transaction` fields.

## Key Changes

- **SimulateItems.res**: Added `svmSimTransaction` type to capture transaction fields from simulate input. Updated parsing to extract and pass these fields alongside existing instruction data.

- **TransactionStore.res & transaction_store.rs**: Extended `SvmTxInput` to include optional transaction fields (signature, allSignatures, feePayer, success, err, fee, computeUnitsConsumed, accountKeys, recentBlockhash, version). Added base58 parsing helpers in Rust to validate and convert these fields.

- **TestIndexer.res**: Fixed block number detection for SVM items — now checks top-level `slot` field in addition to nested `block.slot`, since SVM simulate items place the slot at the top level.

- **Test coverage**: Added `SvmSimulateTransaction_test.res` demonstrating transaction field access in handlers. Moved Metaplex demo from live scenario to template with unit tests (`svm_metaplex_template/typescript/src/indexer.test.ts`), removing dependency on live Solana RPC.

## Implementation Details

- Transaction fields are optional throughout the pipeline to maintain backward compatibility with existing simulate items that don't provide them.
- Base58 parsing in Rust validates address and key formats, returning descriptive errors for invalid input.
- SVM block detection now handles both top-level `slot` (primary) and nested `block.slot` (override) to support the documented simulate item shape.

https://claude.ai/code/session_0172e2LaWQG4EcW6cHDuroQL


packages/cli/templates/static/svm_metaplex_template/typescript/src/indexer.test.ts
```diff
@@ -0,0 +1,104 @@
+import { describe, it } from "vitest";
+import { createTestIndexer } from "envio";
+
+// `start_block` from config.yaml. Simulated instructions are placed by slot and
+// only run when the slot is inside the configured range.
+const START_SLOT = 417_920_000;
```
DZakh: I think the slot can be omitted and it will be derived automatically by the test indexer framework. Should be.

Issue #655, opened 2025-07-22
`error` word support in event signature


DenhamPreen:
<img width="3024" height="1104" alt="Image" src="https://github.com/user-attachments/assets/3bdccc99-d12e-4291-afe4-3f2a6316c202" />

Need to add `error` to reserved words handling

revisedbytom:
I'd like to add a focused regression fixture for an ABI event parameter named `error` and then extend the existing reserved-word handling consistently. Before implementation, should `error` be escaped/renamed using the same generated-identifier convention as the current reserved words, rather than rejected during config validation?

DZakh:
This is fixed

PR #1599, opened 2026-08-31
Test event and parameter named "error"

DZakh:
Regression fixture for https://github.com/enviodev/hyperindex/issues/655: an EVM event named `error` carrying a param named `error`.

Investigation result: the case already works on `main`, so this PR only pins it. `error` is not a ReScript keyword, so `RecordField::to_valid_rescript_name` leaves it untouched, and it does not belong in `RESERVED_NAMES` either — that list covers `Object.prototype` members plus `enum`, names that stay broken after codegen capitalizes them. An event named `error` capitalizes to a `module Error` nested in the contract module, which shadows nothing that matters.

Verified two ways:

- The fixture here (user API rung): config parse, type-check of handler and test source against the generated `types.d.ts`, and end-to-end processing.
- Locally, adding `Error(address indexed to, uint256 error)` to `scenarios/test_codegen` — codegen emits `type params = {to: Address.t, error: bigint}`, and both `rescript` and `tsc --noEmit` pass. Not committed; the fixture above covers it.

https://claude.ai/code/session_01FRHpWvkmdnC6cxxMupuAst

## Summary by CodeRabbit

* **New Features**
  * Added support for the StablesKinshipGrass network (chain ID 988).

* **Tests**
  * Added regression coverage for indexing events named `error` with parameters also named `error`.
  * Verifies generated configuration, schema, event handling, simulated input, and resulting account state.

### DIFF
--- a/packages/cli/src/config_parsing/chain_helpers.rs
+++ b/packages/cli/src/config_parsing/chain_helpers.rs
@@ -385,6 +385,9 @@ pub enum Network {
     #[subenum(HypersyncChain, NetworkWithExplorer)]
     Sophon = 50104,
 
+    #[subenum(HypersyncChain)]
+    StablesKinshipGrass = 988,
+
     StatusSepolia = 1660990954,
 
     #[subenum(HypersyncChain)]
@@ -604,6 +607,7 @@ impl Network {
             | Network::Injective
             | Network::Megaeth
             | Network::SeiTestnet
+            | Network::StablesKinshipGrass
             | Network::StatusSepolia
             | Network::Tempo
             | Network::Tron
--- a/packages/envio-tests/test/ErrorEventName_test.res
+++ b/packages/envio-tests/test/ErrorEventName_test.res
@@ -1,0 +1,56 @@
+// https://github.com/enviodev/hyperindex/issues/655
+let _ = InternalTestIndexer.fromUserApi(
+  ~configYaml=`
+name: error-event
+chains:
+  - id: 1
+    start_block: 0
+    contracts:
+      - name: Token
+        address: "0x1111111111111111111111111111111111111111"
+        events:
+          - event: error(address indexed to, uint256 error)
+`,
+  ~schema=`
+type Account {
+  id: ID!
+  balance: BigInt!
+}
+`,
+  ~handlers=`
+import { indexer } from "envio";
+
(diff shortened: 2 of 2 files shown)


DZakh:
`hypersync-health-check` is failing on this head, and the failure is not this PR's:

```
HyperSync API has chains absent from the Network enum:
    #[subenum(HypersyncChain)]
    StablesKinshipGrass = 988,
```

That check compares the live HyperSync API's chain list against the `Network` enum, so it turns red whenever HyperSync adds a chain — independently of any diff. This PR adds one ReScript test file (`packages/envio-tests/test/ErrorEventName_test.res`) and touches no Rust. I reproduced the identical panic locally on a tree whose Rust code is exactly the base commit (`72ed729`), which is the same commit whose `main` run was green on Aug 28 — the drift appeared upstream since then. Everything else on this head is green or still running (`cargo-test`, `build-envio-package`, `scenarios-test` passed).

No fix for it exists to port, and the fix is a product change unrelated to this PR (adding a supported network), so I'm not widening the PR with it. The patch, for whoever picks it up — in `packages/cli/src/config_parsing/chain_helpers.rs`, in the alphabetical run of `Network` variants (between `Sophon` and `StatusSepolia`):

```rust
    #[subenum(HypersyncChain)]
    StablesKinshipGrass = 988,
```

plus the corresponding entry in the network classification match further down the same file, if the chain belongs there. `cargo test --features hypersync_health --test hypersync_health` verifies it.

Re-running the failed job once to confirm it is not transient.

---

DZakh:
Fixed in `d288358` — at the repo owner's request I've taken the health-check failure into this PR rather than leaving it for a separate one.

`stables-kinship-grass` (chain id 988, tier SILVER) is now in the `Network` enum as a `HypersyncChain`, placed to keep the alphabetical-order test happy, plus the matching arm in `get_max_reorg_depth` (unknown depth, like the other recently added chains). Nothing else was needed: the HyperSync URL is derived from the chain id, and the chain has no explorer or Graph network, so `CommandLineHelp.md` is unchanged.

Verified locally: `cargo test --features hypersync_health --test hypersync_health` now passes (it probes the new endpoint too) — it failed with the same panic before the change; `cargo test -p envio` is green (541 passed), as are `cargo fmt --check` and `clippy`.

---

PR #1602, opened 2026-08-31
Support start_block: latest

DenhamPreen:
## Summary
- Adds `start_block: latest` as an alternative to a fixed integer in `config.yaml`
- On first deploy it resolves once to the chain's current head and persists that block; restarts reuse the persisted value instead of re-resolving, so downtime gets backfilled rather than skipped
- Works for EVM, Fuel, and SVM chains; the test indexer treats an unresolved `latest` as unconstrained (0)

## Test plan
- [x] `cargo test --lib` in `packages/cli` (549 passed, includes 8 new tests)
- [x] `pnpm vitest run` in `packages/envio-tests` (1227 passed, includes 4 new tests covering config parsing, first-deploy resolution, restart not re-resolving, and the end_block conflict error)

https://claude.ai/code/session_018z88cNc14mddGWAPg8GXUG

## Summary by CodeRabbit

- **New Features**
  - Configure chain start blocks using a fixed block number or `latest`.
  - `latest` resolves from the chain head and persists across normal restarts, while full resets resolve it again.
  - Configure EVM and Fuel byte values as hexadecimal strings or `Uint8Array`.
  - SVM instruction discriminators now support hexadecimal prefixes of any whole-byte length.

- **Bug Fixes**
  - Improved validation for start blocks, contract start blocks, end blocks, and chain reorganizations.
  - Improved rollback handling and progress tracking across multiple chains.

- **Breaking Changes**
  - Removed SVM instruction account-filter and inner-instruction configuration options.


packages/envio/evm.schema.json
```diff
@@ -483,10 +483,8 @@
           "minimum": 0
         },
         "start_block": {
-          "description": "The block at which the indexer should start ingesting data",
-          "type": "integer",
-          "format": "uint64",
-          "minimum": 0
+          "description": "The block at which the indexer should start ingesting data, or \"latest\" to start from the chain's current head block when the indexer is first deployed. Once resolved, the concrete block is persisted and reused every time the indexer resumes normally (for example recovering from a crash), so downtime is backfilled instead of skipped. Running `envio start`/`dev` with -r (--restart) resets this like any other config change: \"latest\" resolves again, against the head at that time.",
```
DenhamPreen, line 497: lmk if you think this is too verbose, I wanted to explicitely differentiate a restart from the cli restart

DenhamPreen:
This doesn't work on per contract start blocks, to implement that the per contract startblock would need to be persisted which it currently isn't.

DZakh:
I think per-contract startblock is not an issue. In general the PR looks good, but I want anoter PR merged before this one. After this, I'll take your branch and make sure it's in main 👍

PR #1608, opened 2026-09-01
Decode SVM instruction args to bigint for 64-bit-plus integers

DZakh:
## Overview

`u64`/`u128`/`i64`/`i128` instruction args reached SVM handlers as decimal strings — the upstream `hypersync-client-solana` decoder deliberately renders them that way in JSON and leaves the bigint conversion to the consumer. Since every SPL token amount is a `u64`, storing any Solana amount into a `BigInt!` entity field required a manual `BigInt(instruction.args.amount)`. EVM and Fuel already deliver wide integers as `bigint`; SVM now matches.

## What changed

- **Shared `ParamValue` (`packages/cli/src/param_value.rs`)**: the EVM decoder's napi value tree, lifted out of `evm_hypersync_source/types.rs` and extended with `Num`/`Null` variants (SVM needs both; EVM never constructs them). The `BigInt` variant now carries plain `{sign_bit, words}` data instead of a napi handle, so the whole tree derives `Debug`/`Clone`/`PartialEq` and Rust unit tests assert full decoded trees directly; the napi `bigint` is built only in `ToNapiValue`.
- **Conversion lives in Rust** (`svm_hypersync_source/borsh_decoder.rs`): after the upstream decode, `args_to_param` re-walks the `serde_json::Value` output against the instruction's Borsh `FieldType` schema — the only place a wide-integer decimal string is distinguishable from a Pubkey or a genuine `string` field. The walk covers `Option` → null, `Vec`/`Array` → arrays, `[u8;32]`/bytes/pubkey → strings, structs, `defined` type references, and Anchor's externally-tagged enums. A shape mismatch is treated like a decode failure (the instruction is dropped, same policy as schema drift).
- **napi boundary**: `EventItem.args_json: String` is gone; the item carries `args: ParamValue` and arrives in JS as a ready value tree. `SvmHyperSyncSource.res` no longer `JSON.parseOrThrow`s per item, and `SimulateItems.res` no longer `JSON.stringify`s user args (which would throw on `bigint`). The registration's *schema* JSON (`argsJson`/`definedTypesJson` arg definitions) intentionally stays a string.
- **Codegen**: `field_type_to_ts_type` maps `u64 | u128 | i64 | i128` → `bigint` (was `string`). All four move together, mirroring EVM/Fuel. Sub-64-bit integers and floats stay `number`; pubkey / bytes / `[u8;32]` stay `string`. Handler args and simulate-item args both derive from the same generated `Programs` table, so both pick it up.

## Breaking change (no docs, per request — this section is the migration note)

SVM handler code that treated wide-integer args as strings must drop the conversion:

```ts
// before
context.Transfer.set({ id, amount: BigInt(instruction.args.amount) });
// after
context.Transfer.set({ id, amount: instruction.args.amount });
```

`BigInt(x)` on an already-`bigint` value is the identity, so tolerant helpers (e.g. `svm_flow_xray`'s `bi()`) keep working; only code relying on string methods or exact `string` typing breaks, and TypeScript flags every site. `rawEvents` is rejected for SVM, so no persisted-JSON path is affected.

## Tests

- **Rust** (`cargo test --lib`, 546 passing): new `borsh_decoder` tests decode real Borsh bytes through `build_program_schema` + `decode_with_schema` and assert full `ParamValue` trees — `u64::MAX`, `u128::MAX`, negative `i64`/`i128` (including a magnitude past 2^64), `Option` none/some, `Vec`, inline struct, and a `defined` enum with unit + struct variants. Existing routing tests updated to assert trees instead of JSON strings; the SVM `.d.ts` insta snapshot captures the `string` → `bigint` flip.
- **envio-tests** (81 SVM/simulate tests passing): `SvmInstructionE2e_test` now feeds `args: { amountIn: 18446744073709551615n, ... }` through simulate and stores `instruction.args.amountIn` straight into a `BigInt!` field — the assignment type-checks only under the new codegen; `SvmInstructionPayload_test` asserts a `u64::MAX` bigint survives payload assembly untouched; `SvmHandlersApi_test` type-equality assertions updated to `bigint`.
- **Live** (`SvmWhereLive_test`, run against solana.hypersync.xyz): every decoded Raydium swap in the pinned window now asserts `typeof args.amountIn === "bigint"` at handler time and that real amounts round-trip into a `BigInt!` entity.
- Both SVM scenarios (`svm_test`, `svm_flow_xray`) re-ran `envio codegen` + `tsc --noEmit` clean.

https://claude.ai/code/session_015x78W5BxVH3PqJ6YU8PSRk

---
_Generated by [Claude Code](https://claude.ai/code/session_015x78W5BxVH3PqJ6YU8PSRk)_

## Summary by CodeRabbit

- **New Features**
  - SVM instruction arguments are now available as decoded values instead of JSON strings.
  - Wide integer arguments are preserved as native `bigint` values.
  - Instructions without declared arguments use an explicit empty-argument representation.

- **Bug Fixes**
  - Improved handling distinguishes undecoded instructions from instructions without schema coverage.
  - Invalid argument selections now produce a clear registration error.
  - Simulation inputs and handler payloads now preserve decoded argument values consistently.

- **Tests**
  - Expanded coverage for decoded arguments, bigint handling, argument selection, and runtime validation.


packages/cli/src/svm_hypersync_source/mod.rs
```diff
@@ -614,9 +615,9 @@ fn build_event_items(
                 accounts: instr.account_arguments.clone(),
                 data: to_hex(&instr.data),
                 is_inner: instr.is_inner,
-                args_json: match &decoded {
-                    Some(args_json) if reg.selects_args => args_json.clone(),
-                    _ => "{}".to_string(),
+                args: match &decoded {
+                    Some(args) if reg.selects_args => args.clone(),
+                    _ => crate::param_value::ParamValue::Obj(vec![]),
```
DZakh: Shouldn't it be null or undefined here? Also, make sure that we skip an instruction if it failed decoding.

DZakh:
## Checked: is `BigDecimal` needed for any SVM arg type?

No. `bigint` is the right terminal type for every numeric Solana arg, and `BigDecimal` belongs one layer up — in the entity field, where it already works today.

**Nothing on the SVM wire is a decimal.** The Borsh grammar upstream exposes exactly: `bool`, `u8`–`i128`, `f32`/`f64`, `string`, `bytes`, `pubkey`, plus the composites. There is no decimal or fixed-point type to map. And the floats do not show up in practice — counting every typed field across the three real protocol IDLs bundled in `scenarios/svm_flow_xray/idls` (Jupiter, Drift, Kamino; ~1,200 fields) turns up **zero** `f32`/`f64`:

| IDL | numeric types present |
| --- | --- |
| jupiter.json | `u64` ×19, `u8` ×12, `u16` ×6, `u32` ×1 |
| drift.json | `u64` ×237, `u16` ×198, `i64` ×167, `u128` ×142, `u32` ×135, `u8` ×120, `i128` ×40, `i8` ×15, `i32` ×7, `i16` ×5 |
| kamino.json | `u64` ×92, `u8` ×54, `u128` ×23, `u16` ×9, `u32` ×4, `i64` ×2 |

**Where a float does appear, `number` is already exactly right.** A Borsh `f64` is an IEEE double, which is precisely what a JS `number` is — the conversion is lossless. Wrapping it in a `BigDecimal` (which is `bignumber.js`, per `packages/envio/index.d.ts`) would imply a precision the binary float never had, at the cost of an allocation per value. `f32` widens into a double exactly too.

**The values that *look* like decimals are scaled integers, and `bigint` is their lossless form.** Drift's IDL carries 142 `u128` and 40 `i128` fields for exactly this reason: `sqrtK` is a `u128`, while `price`, `baseAssetAmount` and `quoteAssetAmount` are `i64` scaled by protocol constants. Same story for Q64.64 `sqrtPrice` in the CLMMs, and for every SPL token amount, whose scale lives in the mint's `decimals` account rather than in the instruction.

The critical point: the scale is program semantics that the IDL does not encode — it literally just says `u128`. Converting eagerly would mean guessing a scale, and a wrong guess corrupts values silently. It would also be genuinely lossy in a way `bigint` is not — Q64.64 fractions need roughly twenty significant digits below the point, so any fixed-precision decimal rounds. Handing over the exact integer is the one choice that cannot be reconstructed after the fact if we get it wrong.

**`BigDecimal` is already available where it is needed.** It is an exported handler type and a supported GraphQL scalar with `precision`/`scale` directives, so a handler does `new BigDecimal(args.sqrtPrice.toString()).div(...)` and stores the result. That is the same division of labor EVM has had all along: Solidity has no decimals either, `uint256` arrives as `bigint`, and the handler applies the token's decimals.

**The only thing that would change this** is an opt-in scale annotation in config (something like `{ name: sqrtPrice, type: u128, scale: 64 }`, or a `q64` alias) that codegen could type as `BigDecimal`. Worth building only if users ask for it — the exact `bigint` stays the correct default underneath it either way.

No change to this PR.

---

packages/cli/src/svm_hypersync_source/mod.rs
DZakh: Good catch — that arm was conflating three different situations, and tracing it turned up a bug. Fixed in fd4f938.

`decode_with_schema` now reports which of the three actually happened:

```rust
pub(crate) enum Decoded { Args(ParamValue), Uncovered, Failed }
```

- **Failed** (schema covers the discriminator, data rejected) → `continue`, instruction dropped. This is what you asked me to make sure of; it already held, and it now also covers a `ParamValue` shape mismatch in the new walk, with `an_instruction_the_schema_cannot_decode_is_dropped` still guarding it.
- **Uncovered** → the bug. A registration declaring neither args nor accounts contributes no schema piece (`has_schema` in `build_schemas`), so its discriminator is absent from the program schema. If a *sibling* instruction in the same program does declare a schema, `schemas.get(program)` hits, upstream returns `UnknownDiscriminator`, and the old code took the `None => continue` path — silently dropping every such instruction, for all its routed registrations. Now it decodes to `{}`, which is exactly what codegen types it as (`args_ts` is `"{}"` when the instruction declares no args). New test `an_instruction_the_program_schema_does_not_cover_keeps_empty_args` fails on the old code with `left: []`.
- **Unselected registration** → your point about the placeholder: it is now `ParamValue::Null`, matching how `logs` already signals unselected, instead of a decoded-looking `{}`. It is never observed either way — `toSvmInstruction` only sets `args` when `hasSelection("args")` — so this is about not having a fake value sitting on the wire.

I kept `{}` rather than null for the *selected-but-nothing-declared* case specifically because the generated type for such an instruction is `{}`; handing back `null` there would break the type contract handlers compile against. Happy to flip it if you'd rather codegen emit `null` for schemaless instructions instead — that would be the consistent alternative, just a wider change.

Full Rust suite green (652), SVM/simulate ReScript tests green (81).

---

DZakh: Done in 8be89b5 — this is better than the null idea, so I dropped that.

An instruction declaring no args now types them as `never`, and `args` drops out of the selectable field names, carrying the reason in the union so the error explains itself:

```
Type '"args"' is not assignable to type '"accounts" | "programId" | "path" | "isInner" | "accountArguments" | "data"
  | "Field 'args' is unavailable: this instruction declares no args in config.yaml, so there is nothing to decode."'
```

Applied at both the `onInstruction` registration (`const F extends SvmFieldsSelectionFor<Programs[P][I]>`) and the standalone `SvmOnInstructionHandler`, so the rule holds wherever a fields selection meets an instruction.

It caught two live cases on the first run: `svm_flow_xray`'s Orca and Meteora `swap` registrations shared a `flowFields` const containing `"args"`, but both are matched on discriminator alone and neither handler ever read the field. They now use an args-free spec.

This also makes the ambiguity you flagged unreachable rather than merely well-defined: a typed handler can no longer reach the empty-object case at all. I kept the runtime `Uncovered` branch as defense for untyped callers, since it is what stops the silent instruction drop.

Green: 652 Rust, 83 SVM/simulate ReScript (two new type tests — the rejection and a positive control that a declaring instruction still selects `args` and reads `bigint`), plus `tsc --noEmit` on both SVM scenarios.

---

DZakh: Reworked per your call in 1ff8c36: the `.d.ts` selection narrowing is gone (`index.d.ts` is back to its original shape), and the check moved to registration time in `HandlerRegister.res`:

```
Invalid "args" field in the fields.instruction option of the "bare" instruction on program "Swapper". The instruction declares no args in config.yaml, so there is nothing to decode. Remove "args" from the selection, or declare the instruction's args.
```

Codegen still types such an instruction's `args` as `never`, so a handler can't read it as an object even if it somehow got that far. The `svm_flow_xray` Orca/Meteora fix stays, since those registrations would now throw on startup.

Tests are runtime now — the rejection plus a positive control that a declaring instruction still registers with `args` selected. 652 Rust, 83 SVM/simulate ReScript, both SVM scenarios `tsc` clean.

---

packages/cli/src/svm_hypersync_source/mod.rs
```diff
@@ -614,9 +618,17 @@ fn build_event_items(
...
-                args_json: match &decoded {
-                    Some(args_json) if reg.selects_args => args_json.clone(),
-                    _ => "{}".to_string(),
+                args: if reg.selects_args {
+                    // A registration that declared no args reads an empty
+                    // object, which is what its generated type promises; one
+                    // whose schema rejected the data never got here.
+                    decoded
+                        .clone()
+                        .unwrap_or(crate::param_value::ParamValue::Obj(vec![]))
```
DZakh: Let's panic since the _or case is unreachable

PR #1622, opened 2026-09-03
Stage ClickHouse rows column by column instead of through a row record

DZakh:
## Why

ClickHouse staging built a throwaway named record for every history row, then hashed each cell by column name on the JS thread. A batch of a few thousand rows paid for that twice: one dict per row, one lookup per cell.

Each registered column now has a reader built once per entity and scope. A row is staged by calling those readers in registration order. Alignment is a name match at build time, then position on every row.

## Scope

- `packages/envio/src/bindings/ClickHouse.res` builds `setCells` / `deleteCells` in `table.columns` order. `clickHouseFieldSchema` is shared by those readers and `makeClickHouseEntitySchema`. It keeps the main-line Bytea passthrough (`S.json(~validate=false)`, not `fieldSchema`) and the Json mapping. `resume` still passes `historyTables`, `replicated`, and `databaseEngine`.
- `makeConverters` and `fillBuilders` are private. Converter build runs inside the `Persistence.StorageError` try.
- `packages/envio-tests/test/lib_tests/ClickHouseStaging_test.res` registers a real table, mocks `stage`, and pins column names plus the `columnValuesInput` list. The fixture covers snake_case, a linked `owner: User!`, Bytes scalars and `[Bytes!]!`, SET/DELETE, and the chain-id tag.
- `ColumnNameFormat_test.res` no longer serializes through `makeSetUpdateSchema`. That stack is not the ClickHouse write path.

Out of scope: changing ClickHouse DDL, resume, or the checkpoint column list.

## Tradeoffs

The leftover-field check still exists for a table that `registerEntityTable` cannot produce. It is cheap, private, and wrapped as `StorageError`. Unreachable from user YAML, so it is not a user-API test.

## Blast Radius

ClickHouse-backed indexers only. Handlers, YAML, and schema stay the same. Rows should match what the old record path wrote. A sloppy rebase onto `#1620` / `#1609` would have dropped Bytes or catalog-less resume. This branch is rebased onto current main and keeps both.

## Verification

```
cd packages/envio-tests && pnpm rescript && pnpm vitest run \
  test/lib_tests/ClickHouseStaging_test.res \
  test/lib_tests/ColumnNameFormat_test.res \
  test/lib_tests/ClickHouse_test.res \
  test/lib_tests/ClickHouseColumnKind_test.res \
  test/ConfigCrossChain_test.res
```

53 tests passed (20 in the ClickHouse lib files, 33 in ConfigCrossChain). Postgres on 5433 is only for vitest global setup. Staging itself never talks to a server.

## Summary by CodeRabbit

- **New Features**
  - ClickHouse staging now supports binary (`Bytea`) column values.
  - Staged entity changes preserve column ordering, typed values, null markers, metadata, checkpoints, and change types.
  - Set and delete operations now handle entity IDs, chain IDs, serialized values, and deleted values consistently.

- **Bug Fixes**
  - Improved validation for mismatched entity fields and database columns.
  - Improved handling of table-specific data conversions during staging.

### DIFF
--- a/packages/envio-tests/test/lib_tests/ClickHouseStaging_test.res
+++ b/packages/envio-tests/test/lib_tests/ClickHouseStaging_test.res
@@ -1,18 +1,14 @@
 open Vitest
 
-// A batch is staged column by column, in the order the table registered them,
-// and each column's values are read straight off the change rather than out
-// of a serialized row. The whole staged payload is pinned at once: a column
-// reading another's values, or a row shape reading the wrong one, shows up as
-// the payload differing rather than as a store holding the wrong data.
-//
-// Registration never reaches the server, so none of this needs a ClickHouse.
-
 let config = InternalTestIndexer.fromUserApi(
   ~schema=`
 enum Kind {
   A
   B
+}
+
+type User {
+  id: ID!
 }
 
 type Row {
@@ -25,10 +21,15 @@ type Row {
   tags: [String!]!
   optKind: Kind
   doc: Json!
+  owner: User!
+  blob: Bytes!
+  optBlob: Bytes
+  chunks: [Bytes!]!
 }
 `,
   ~configYaml=`
 name: clickhouse-staging
+bytes_type: uint8array
 disable_default_cross_chain: true
 storage:
   postgres:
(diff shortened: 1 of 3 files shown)


DZakh:
CI note: `envio-tests (postgres)` failed on the first run in `test/lib_tests/SourceManager_test.res` — "After primary fails in executeQuery, waits for recovery timeout before retrying it" ("Expected one pending call to fallbackMock"). That is a timer-based source-failover test; this PR changes only `ClickHouse.res` and adds a ClickHouse staging test, and the ClickHouse leg of the same run (which executes the same file) passed. The test passes 5/5 locally on this branch. Re-running the failed job once; if it fails again I'll treat it as this PR's.

---

PR #1538, opened 2026-08-07
Envio Subgraph (experimental)

DZakh:
Adds support for running unmodified subgraph projects on HyperIndex by translating `subgraph.yaml` manifests and GraphQL schemas into envio configs, and providing a graph-ts compatibility shim that runs the AssemblyScript mappings unchanged.

## Key Changes

**Manifest and Schema Translation**
- `packages/cli/src/subgraph/manifest.rs`: Parses and validates `subgraph.yaml`, enforcing strict conformance to known features and versions
- `packages/cli/src/subgraph/schema.rs`: Translates GraphQL schemas from subgraph format to envio format, mapping scalar types and handling directives
- `packages/cli/src/subgraph/abi.rs`: Resolves event signatures against contract ABIs
- `packages/cli/src/subgraph/usage.rs`: Analyzes mapping source to determine which `event.transaction` and `event.block` fields are actually read, avoiding unnecessary RPC calls
- `packages/cli/src/subgraph/errors.rs`: Collects and reports all validation errors in a single pass

**graph-ts Compatibility Shim**
- `packages/envio/src/subgraph/graph-ts.ts`: Complete implementation of `@graphprotocol/graph-ts` API surface, including `ByteArray`, `Address`, `BigInt`, `BigDecimal`, entity classes, and store operations
- `packages/envio/src/subgraph/calls.ts`: Contract call execution as envio effects, evaluated at the event's block
- `packages/envio/src/subgraph/blocks.ts`: Block timestamp fetching via HyperSync or RPC fallback
- `packages/envio/src/subgraph/hosts.ts`: Host operations (IPFS, Arweave, ENS, balance/code checks)
- `packages/envio/src/subgraph/division.ts`: AssemblyScript-compatible integer division
- `packages/envio/src/subgraph/conformance.ts`: Type conformance checking against real graph-ts package
- `packages/envio/src/subgraph/runtime.ts`: Handler registration and execution wrapper
- `packages/envio/src/subgraph/scope.ts`: Async context tracking for handler mode and schema information
- `packages/envio/src/subgraph/errors.ts`: Error messages for unsupported features and refused operations

**Type Definitions**
- `packages/envio/src/subgraph/graph-ts-types/`: Complete TypeScript declarations for graph-ts covering ethereum, cosmos, near, arweave, and starknet chains
- `packages/envio/scripts/generate-graph-ts-types.mjs`: Script to generate type declarations from the real graph-ts package

**Integration**
[...]

PR #1626, opened 2026-09-04
Envio Subgraph (experimental)

DZakh:
Supersedes #1538, which should be closed in favour of this one. This branch forked from #1538 and now contains all of its work plus a newer merge of `main`; #1538's branch has no commits this one lacks.

`envio dev` / `envio start` work inside an unmodified subgraph project — `subgraph.yaml` + `schema.graphql` + AssemblyScript mappings + `abis/`, no `config.yaml` and nothing envio-shaped in the tree. The CLI detects `subgraph.yaml`, translates the manifest and schema into an envio config, and points the handler entry at a runtime that imports the mappings unchanged. Only `@graphprotocol/graph-ts` is shimmed; the project's own `generated/` — real `graph codegen` output — executes as-is on top of it.

Design and the compatibility matrix are in `docs/subgraph-compat.md`.

## How it fits together

- **Translator** (`packages/cli/src/subgraph/`) — manifest and schema into an envio config, strict about what it accepts, reporting every finding at once rather than the first. Reachable without the filesystem through `from_subgraph` for tests.
- **Runtime** (`packages/envio/src/subgraph/`, never exported) — each manifest handler becomes an `indexer.onEvent` / `onBlock` / `contractRegister` wrapper. Ambient APIs (`entity.save()`, `dataSource.address()`) find their envio context through `AsyncLocalStorage`.
- **Sync bridge** (`UserContext.res`) — mappings are synchronous and envio's store is not. A read that misses memory suspends the handler, schedules the load and replays it. Internal, reached by the runtime through a cast; no public API is committed to here.
- **graph-ts shim** — type-checked against the real `@graphprotocol/graph-ts` surface on every CI run, with each known gap recorded in `conformance.ts`.

## Two pins the translator makes, not overridable

- `address_format: lowercase` — graph-ts renders addresses lowercase, and id/derived-key parity depends on it.
- `bytes_type: uint8array` — graph-ts `Bytes` *is* a `Uint8Array`, so a `Bytes` column stores the bytes rather than a hex rendering of them. Ids stay text, since envio ids are strings.

## Where it diverges from graph-node

- `@entity(immutable: true)` is validated, then dropped — the write-once check is not enforced. Any subgraph that runs cleanly on graph-node never trips it.
- Every `@derivedFrom` field is emitted as `[T!]!` whatever nullability it was written with, including the one-to-one form. The lookup is the same either way, and `graph codegen` reads it back as an array however it is declared.
- Anything unsupported fails with one uniform, loud error — never a silent no-op.

Query-API differences (Hasura vs graph-node GraphQL) are out of scope.

## Tests

Three rungs, all run in CI:

- `packages/envio-tests/test/Subgraph*` — real manifest, schema and mapping source through `fromSubgraph`, executed. Covers the graph-ts surface, the sync bridge, field types, contract calls, derived fields, and the refusals.
- `scenarios/subgraph_test` — a whole subgraph project with `generated/` checked in as real `graph codegen` output, so the mappings run against the same bytes a subgraph developer gets. A separate CI job regenerates the goldens and fails on drift.
- `cargo test subgraph::` — 52 translator cases.

https://claude.ai/code/session_01DBgCCjPmPvx6C2UcTWiquc

---
_Generated by [Claude Code](https://claude.ai/code/session_01DBgCCjPmPvx6C2UcTWiquc)_

## Summary by CodeRabbit

* **New Features**
  * Added support for running unmodified Graph subgraphs on HyperIndex.
  * Added manifest and schema translation, Graph TypeScript compatibility, contract calls, block handlers, templates, and host integrations.
  * Added Hyperliquid support as a Graph-compatible network.
  * Added clear diagnostics for unsupported features, unknown fields, missing RPC endpoints, and code-generation issues.
  * Added synchronous mapping operations with suspend-and-replay behavior.

* **Documentation**
  * Added compatibility and setup guidance for subgraph projects.

* **Tests**
  * Added comprehensive integration and compatibility coverage for subgraph indexing and runtime behavior.


DZakh:
alpha-10 release here

PR #1640, opened 2026-09-10
Expose Solana address-lookup-table resolved addresses

JasoonS:
### Expose Solana address-lookup-table resolved addresses

Solana v0 (versioned) transactions reference most of their accounts indirectly through Address Lookup Tables. HyperIndex exposes only the static `accountKeys`, so a handler doing account analysis silently sees a fraction of the transaction. It is silent, which is the dangerous part: you get a plausible, wrong answer with no error.

Measured on Kamino: 99% of transactions are versioned, with roughly 11 static keys against 23 ALT-resolved keys each. In the transaction the new live test pins, the indexed Kamino program is **not in `accountKeys` at all**. It arrives via a lookup table:

```
static=11   ALT-writable=11   ALT-readonly=11   total=33
KLend in accountKeys?   false
KLend in ALT-readonly?  true
```

An indexer reading only static keys cannot find the very program it is indexing in that transaction's account list.

HyperSync already serves `loaded_addresses_writable` and `loaded_addresses_readonly`, and `hypersync-client-solana` already decodes them into `simple_types::Transaction`. So this is purely additive plumbing along the path `accountKeys` already walks. No dependency bump.

**Commit 1** adds `loadedAddressesWritable` and `loadedAddressesReadonly`: two flat fields, one per HyperSync column, one selection-mask bit each. New `SvmTxField` ordinals 12 and 13 are appended; nothing is renumbered.

**Commit 2** adds `allAccountKeys`, derived in the store as `accountKeys ++ loadedAddressesWritable ++ loadedAddressesReadonly`. That is Solana's own account-resolution order and what a transaction's account indexes point into, so it is what most consumers actually want. Split out so commit 1 can land independently. A legacy transaction with no ALTs still materialises exactly its static keys.

No config-schema change is needed: `fields.transaction` names are validated at runtime in `EventConfigBuilder.res`, not enumerated in `svm.schema.json`. `make update-schemas` produces no diff.

### Testing

- `cargo test --lib`: 782 pass. New coverage at the selection layer (`loaded_addresses_fetch_the_lookup_table_columns`), the decode layer (`svm_loaded_addresses_decode_as_separate_base58_lists`), and the concatenation order (`svm_all_account_keys_concatenate_in_resolution_order`).
- The existing ReScript/Rust field-name and ordinal pinning tests were extended rather than relaxed.
- Full `packages/envio-tests` suite: 162 files, 1,579 tests, 0 failures.
- New `SvmLoadedAddressesLive_test.res` runs a real indexer against `solana.hypersync.xyz` and asserts real ALT addresses reach a handler, including the program-id case above. It follows the existing `SvmBytesLive` / `SvmWhereLive` pattern and uses the `ENVIO_API_TOKEN` the workflow already provides.
- `cargo fmt`, `cargo clippy --all-targets --all-features` and `pnpm lint` are clean.

### Not covered

- No `scenarios/` codegen or Postgres-backed run. The proof is at the user-API rung, which does exercise config through generated types, handler and entity via the real Rust store.
- Built and run on `darwin-arm64` only; CI is the first Linux exercise.
- Docs site not updated.

https://claude.ai/code/session_016vsg7oJZQ5fgkM73tJ7dvg

## Summary by CodeRabbit

* **Bug Fixes**
  * SVM transaction `accountKeys` now includes addresses loaded through writable and readonly address lookup tables.
  * Account keys are returned in Solana’s expected order, while legacy transactions continue to work as before.

* **Documentation**
  * Clarified that `accountKeys` contains static keys followed by writable and readonly lookup-table addresses.

* **Tests**
  * Added coverage for versioned transactions and live resolution of address lookup-table keys.


DZakh:
I think in the PR we should only change `accountKeys` to be what you made `allAccountKeys` like. This matches `jsonParsed` format and I assume what users would expect from `accountKeys` field. You access a key by index and it just works without needing to know that there's `allAccountKeys`. If somebody wants to know writable/readable/static details, we can add them as separate fields later. Let me know what you think

JasoonS:
Heads-up: I've published an alpha from current main so we can run this on a Kamino indexer before the next release: `envio@3.11.0-svm-alpha.1` (tag `v3.11.0-svm-alpha.1` on `846a23f`), on the npm `dev` tag only. `latest` and `next` are untouched. It includes everything unreleased on main since 3.11.0, not just this PR. One side effect: `dev` previously pointed at `3.10.0-resolvers.1`.

